PR #206, opened 2019-11-24
replace UI installer section

### LINKED ISSUE #186: Remove unsupported GUI installer section
We should remove the GUI installer from the docs, it's outdated and not supported anymore.
https://docs.kubermatic.io/installation/install_kubermatic/_installer/

Maybe we could let then entry stay and just write a hint, that we are working on a Kubermatic Operator and a new version of the installer

toschneck:
closes #186

Instead of the not maintained UI installer, add an hint that the operator will get relased soon.


xrstf:
@toschneck can you please rebase this so it has a valid .prow.yaml?

xrstf:
My other PR has already reworked the docs for the operator, so this can be closed.

PR #213, opened 2019-12-02
Added tutorial on node deployments

lindtdev:
This PR adds a tutorial on adding and editing node deployments.


kdomanski:
There should be a copy of this in the `master` folder as well.

lindtdev:
@kdomanski I copied both tutorial 7 and tutorial 8 to master (not tutorial 6 because you opened a separate PR for it)

PR #227, opened 2020-01-27
Replace initech with example name

### LINKED ISSUE #163: Change all Domains names initech.com to example.com
Don't use real names in our docs. initech.com is a company use example.com instead

scheeles:
Close #163


xrstf:
RIP Initech :-(

/lgtm

PR #232, opened 2020-02-07
fix OWNERS_ALIASES

xrstf:
Bad YAML makes me sad. Even though I originally approved it.


xrstf:
Aliases are broken, only people named directly can approve.

PR #240, opened 2020-02-21
Add Google Compute API enabling

themue:
Signed-off-by: Frank Mueller <mail@themue.dev>


xrstf:
Why are we updating old releases? I thought the plan was to only ever put new documentation into the master version, at most maybe in the current release. Yet time and time again we have _some_ PRs that just update the old docs as well.

/approve

PR #244, opened 2020-02-25
update preset doc

xrstf:
This updates the presets section for the Kubermatic Operator.


toschneck:
@xrstf should be also added for the `2.13` branch or?

PR #246, opened 2020-02-25
update addon doc

xrstf:
This updates the addon documentation. The list of default addons was outdated and there was no mention of how the Kubermatic Operator does the configuration nowadays.

This also fixes #180 by explaining the injected template variables.

PR #250, opened 2020-03-11
Fix addons documentation

chrkl:
Added missing colon in Helm values and fixed key for default addons.


xrstf:
For the master version, this is superseded by #246.

toschneck:
@chrkl I think we should place this in the `2.13` docs.

chrkl:
@xrstf modified to 2.12 and 2.13 only. Please review

PR #268, opened 2020-04-01
document user settings

kgroschoff:
Document the user settings feature in master and v2.13


content/kubermatic/master/advanced/user_settings/_index.en.md
```diff
@@ -0,0 +1,27 @@
...
+
+### User Settings Overview
+
+![User Settings view](/img/advanced/user_settings/view.png)
+
+The user settings view is devided into two columns: 
+The left side contains all account details, such as the name and the email of the user, whereas the right side lists all possible settings that a user can make. 
+
+* `Items per page` will change the number of visible items per page in a table. It will be applied to all tables within kubermatic.
+* `Theme` lists all available themes for the dashboard so the user can select its prefered one.
```
kgroschoff, line 23: I think we could add a reference to the theme documentation (as soon as it exists).
Btw: Theming as a new topic or within https://docs.loodse.com/kubermatic/master/advanced/custom_ui/ ? WDYT?

maciaszczykm: We also have the theming docs in dashboard's repo, it would be good if someone would move it here. We can use the same topic if you want.

PR #310, opened 2020-05-15
Added script to run Cypress

lindtdev:
This is the script to run the tests in Cypress.


lindtdev:
I wanted to fix this skript in this PR but decided to move it to the other e2e-related PR where it can be reviewed in context.

PR #335, opened 2020-06-12
fix: remove "-m" as kubeone install flag

domi-nik-:
"-m" is not supported by kubeone install but used in documentation


xmudrii:
@domi-nik- Docs on the website are targeting the `master` branch. If you're using v0.11, the `-m` flag is not going to work, since it's not implemented in that release.

In the case you're using v0.11, you can also check docs in [our GitHub repo](https://github.com/kubermatic/kubeone/tree/release/v0.11/docs).

domi-nik-:
OK, I see. Thank you!

PR #356, opened 2020-07-03
fixes of missing config in EE version

toschneck:
- image name for ee version
- missing certificate issuer spec for kubermatic configuration

Discovered during test installation with latest images:
```
quay.io/kubermatic/dashboard-ee:e4a7535bf0253fb6d5a397a72e35c5d7d4591658
quay.io/kubermatic/kubermatic-ee:2acfea59bc2edb09ccbecf119b0d99e5c7e95538
```

### DIFF
--- a/content/kubermatic/master/installation/install_kubermatic/_index.en.md
+++ b/content/kubermatic/master/installation/install_kubermatic/_index.en.md
@@ -143,6 +143,9 @@ dex:
     userID: "08a8684b-db88-4b73-90a9-3cd1661f5466"
 
 kubermaticOperator:
+  image:
+    # image id for EE version
+    repository: "quay.io/kubermatic/kubermatic-ee"
   # insert the Docker authentication JSON provided by Loodse here
   imagePullSecret: |
     {
@@ -252,6 +255,10 @@ spec:
     # this domain must match what you configured as dex.ingress.host
     # in the values.yaml
     domain: kubermatic.example.com
+    # issuer of `nginx-ingress-controller` 
+    # execute: kubectl get clusterissuers
+    certificateIssuer:
+      name: letsencrypt-prod
 
   # These secret keys configure the way components commmunicate with Dex.
   auth:
--- a/content/kubermatic/v2.14/installation/install_kubermatic/_index.en.md
+++ b/content/kubermatic/v2.14/installation/install_kubermatic/_index.en.md
@@ -141,6 +141,9 @@ dex:
     userID: "08a8684b-db88-4b73-90a9-3cd1661f5466"
 
 kubermaticOperator:
+  image:
+    # image id for EE version
+    repository: "quay.io/kubermatic/kubermatic-ee"
   # insert the Docker authentication JSON provided by Loodse here
   imagePullSecret: |
     {
@@ -247,6 +250,10 @@ spec:
     # this domain must match what you configured as dex.ingress.host
     # in the values.yaml
     domain: kubermatic.example.com
+    # issuer of `nginx-ingress-controller` 
+    # execute: kubectl get clusterissuers
+    certificateIssuer:
+      name: letsencrypt-prod
 
   # These secret keys configure the way components commmunicate with Dex.
   auth:


xrstf:
If this happened in your installation, it's a bug in the operator (you probably. The EE operator should default to the EE repository on quay: https://github.com/kubermatic/kubermatic/blob/release/v2.14/api/pkg/resources/resources_ee.go#L24-L31

If you want to debug this further, make sure the KubermaticConfiguration in the cluster *does not contain* explicit `dockerRepository` settings and run the operator with `-log-debug`. The operator will log the defaults it applies to the configuration on every reconcile loop. It should say that it defaulted to `quay.io/kubermatic/kubermatic-ee`. Also the operator logs it's edition and version when it boots, so check the first few lines of output if you actually have an EE operator.

The fix for the certIssuer however is real and very welcome :-)

xrstf:
I checked the Docker image you quoted above:

```
$ docker run --rm -it quay.io/kubermatic/kubermatic-ee:2acfea59bc2edb09ccbecf119b0d99e5c7e95538 kubermatic-operator -namespace foo
{"level":"info","time":"2020-07-03T17:41:01.454Z","caller":"kubermatic-operator/main.go:91","msg":"Moin, moin, I'm the Kubermatic Enterprise Edition Operator and these are the versions I work with.","kubermatic":"2acfea59bc2edb09ccbecf119b0d99e5c7e95538","ui":"e4a7535bf0253fb6d5a397a72e35c5d7d4591658"}
```

That's fine.

```
$ kube-dev
$ k -n kubermatic logs -f kubermatic-operator-6dcfccccbf-8d6sb
....
{"level":"debug","time":"2020-07-03T17:40:06.392Z","logger":"kubermatic-master-operator","caller":"common/defaults.go:585","msg":"Defaulting Docker repository","config":"kubermatic/kubermatic","field":"api.dockerRepository","value":"quay.io/kubermatic/kubermatic-ee"}
{"level":"debug","time":"2020-07-03T17:40:06.392Z","logger":"kubermatic-master-operator","caller":"common/defaults.go:585","msg":"Defaulting Docker repository","config":"kubermatic/kubermatic","field":"ui.dockerRepository","value":"quay.io/kubermatic/dashboard-ee"}
....
```

Dev uses the same Docker image (well, the same repository, but two commits further in the future) and enabling debug logging reveals that the defaulting also works.

Checking the `kubermatic-api` Deployment also shows that the `kubermatic/kubermatic-ee` repo is used.

TL;DR: Works as intended on dev.

content/kubermatic/master/installation/install_kubermatic/_index.en.md
```diff
@@ -143,6 +143,9 @@ dex:
     userID: "08a8684b-db88-4b73-90a9-3cd1661f5466"
 
 kubermaticOperator:
+  image:
+    # image id for EE version
+    repository: "quay.io/kubermatic/kubermatic-ee"
```
xrstf, line 148: So if I follow the installation documentation like this, I cannot install the CE version. Can we make it more clear that this is an option **if** the user wants the EE?

toschneck: how does the config look like for the CE version? 
is it `docker pull quay.io/kubermatic/kubermatic` as default?

xrstf: Yep, for the CE version you don't need to set `kubermaticOperator.image.repository` at all, as the default value already points to the CE repo.

xrstf:
I think this can be closed.

xrstf:
The CE/EE archives we now offer already preconfigure the operator now to use the correct Docker repository. We also ship example configs for CE/EE (see https://github.com/kubermatic/kubermatic/blob/master/charts/values.example.ce.yaml and https://github.com/kubermatic/kubermatic/blob/master/charts/values.example.ee.yaml), so this can be closed.

@toschneck

toschneck:
agree

PR #359, opened 2020-07-06
Remove nodeport-proxy helm chart

youssefazrak:
Currently, when installing nodeport-proxy following the documentation, nodeport-proxy gets deployed twice: by the operator, and later when installing the helm chart. 

With AWS, this leads to issues as the "helm chart" nodeport-proxy LB doesn't get assigned an external IP, and coupled with https://github.com/kubermatic/kubermatic/pull/5609, the communication between the userclusters and control planes are not possible.

As we are moving to the operator install, I've removed the whole nodeport-proxy helm chart section so we get only one clean installation.

### DIFF
--- a/content/kubermatic/master/installation/add_seed_cluster/_index.en.md
+++ b/content/kubermatic/master/installation/add_seed_cluster/_index.en.md
@@ -162,12 +162,12 @@ on the seed.
 #### With LoadBalancers
 
 When your cloud provider supports Load Balancers, you can find the target IP / hostname by looking at the
-`nodeport-lb` Service:
+`nodeport-proxy` Service:
 
 ```bash
-kubectl -n nodeport-proxy get services
+kubectl -n kubermatic get services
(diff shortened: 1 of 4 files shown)


toschneck:
@youssefazrak is this  for 2.14.x and everyhting above or only for the current master build/weekly releases?

youssefazrak:
@toschneck should be for every version using the operator. IIRC 2.14 and later.

toschneck:
@youssefazrak ok :+1: then you would need to adapt also the `/master/content/kubermatic/v2.14/` description :wink:

youssefazrak:
@toschneck nicely spotted ;)
Indeed this is taken care of by the Operator. I will fix this part of the documentation.

youssefazrak:
@toschneck it has been fixed. Let me know if there is anything else.

toschneck:
@youssefazrak is this closed due to another PR ?

youssefazrak:
@toschneck yes, this one #361

PR #360, opened 2020-07-08
improve minio doc

toschneck:
- add missing access values definition for minio chart
- rename storage class to a kubermatic name schema
- add some more explanations
- fix s3-exporter namespace for 2.14+


toschneck:
@xrstf @youssefazrak please review again

xrstf (approved):
/override pre-docs-spellcheck
/approve

Overridden because the test failed due to an upstream bug.

PR #402, opened 2020-08-05
Remove experimental warnings from seed and kubermaticconfiguration docs

irozzo-1A:
Remove experimental warnings from seed and kubermaticconfiguration docs starting from version 2.14 and update example manifests.

### DIFF
--- a/content/kubermatic/v2.14/data/kubermaticConfiguration.yaml
+++ b/content/kubermatic/v2.14/data/kubermaticConfiguration.yaml
@@ -8,11 +8,12 @@ spec:
   api:
     # AccessibleAddons is a list of addons that should be enabled in the API.
     accessibleAddons:
-    - node-exporter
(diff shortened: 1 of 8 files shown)


content/kubermatic/v2.14/data/kubermaticConfiguration.yaml
```diff
@@ -8,11 +8,12 @@ spec:
   api:
     # AccessibleAddons is a list of addons that should be enabled in the API.
     accessibleAddons:
-    - node-exporter
+      - node-exporter
+      - gatekeeper
```
youssefazrak, line 12: This needs to be applied in the master version too

youssefazrak: Related to the master kubermaticConfiguration.yaml file. Applies to all changes below.

PR #438, opened 2020-08-28
document migration of helm componentOverride settings

furkhat:


irozzo-1A:
Closing at the moment as we did not decide yet if this will be the final solution.

PR #439, opened 2020-08-28
fix typo in addon documentation

erdii:
spec.userCluster.addons instead of spec.userClusters.addons


erdii:
naive question: shouldn't the schema be typechecked against the CRD?

we lost some hours due to overseeing this typo

moelsayed:
@erdii I think typechecking and validation has to be defined in the CRD itself. 

/lgtm
/approve

PR #441, opened 2020-09-04
Clean up naming between Company and KKP

fairhaven:


KrissiWi (changes requested):
@fairhaven : Please make sure that we have a space before the braket, not `Kubermatic Kubernetes Plattform(KKP)` but `Kubermatic Kubernetes Platform (KKP)`

PR #458, opened 2020-10-10
update latest permissions at vsphere

toschneck:
- add smaller missing permissions
- clarifaction what level network permissions need to get applied
- add link to terraform setup


xrstf (approved):
/approve

PR #460, opened 2020-10-13
Add spell checker job in CI/CD

irozzo-1A:
Add spell checker job


.prow.yaml
```diff
@@ -1,4 +1,20 @@
 presubmits:
+  - name: pre-kubermatic-spellcheck
```
xrstf, line 2: The job should be called `pre-docs-spellcheck`.

data/kubermatic/master/runbook.json
```diff
@@ -887,7 +887,7 @@
         {
           "alert": "PromBadConfig",
           "annotations": {
-            "mesage": "Prometheus failed to reload config.",
+            "message": "Prometheus failed to reload config.",
```
xrstf, line 890: NB: This must be fixed upstream (Kubermatic repo), or else the next sync will overwrite this fix.

xrstf (changes requested):
One change request, one hint, otherwise very nice.

data/kubermatic/master/runbook.json
irozzo-1A: I'm going to add the same job in kubermatic repo anyway

xrstf (approved):
/approve

PR #461, opened 2020-10-13
Add datastore and datastore cluster docs for vSphere

irozzo-1A:
* Align seed manifests
* Add datastore and datastore cluster documentation

### DIFF
--- a/content/kubermatic/master/requirements/cloud_provider/_vsphere.en.md
+++ b/content/kubermatic/master/requirements/cloud_provider/_vsphere.en.md
@@ -193,3 +193,40 @@ Upstream Kubernetes has been working on the issue for a long time now and tracki
 * <https://github.com/kubernetes/kubernetes/issues/67900>
 * <https://github.com/kubernetes/kubernetes/issues/71829>
 * <https://github.com/kubernetes/kubernetes/issues/75342>
+
+### Datastores and Datastore Clusters
+
+*Datastore* in VMWare vSphere is an abstraction for storage.
+*Datastore Cluster* is a collection of datastores with shared resources and a
+shared management interface.
+
+In KKP *Datastores* are used for two purposes:
+- Storing the VMs files for the worker nodes of vSphere user clusters.
+- Generating the vSphere cloud provider storage configuration for user clusters.
+    In particular to provide the `dafault-datastore` value, that is the default
+    datastore for dynamic volume provisioning.
+
+*Datastore Clusters* can only be used for the first purpose as it cannot be
+specified directly in [vSphere cloud configuration][vsphere-cloud-config].
+
+There are two places where Datastores and Datastore Clusters can be configured
+in KKP
+
+- At datacenter level (either with [Seed CRD]({{< ref "../../concepts/seeds" >}})
+    or [datacenters.yaml]({{< ref "../../concepts/seeds" >}})) is possible to
+    specify the default *Datastore* that will be used for user clusters dynamic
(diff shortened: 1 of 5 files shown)


content/kubermatic/master/requirements/cloud_provider/_vsphere.en.md
```diff
@@ -193,3 +193,40 @@ Upstream Kubernetes has been working on the issue for a long time now and tracki
...
+    datastore for dynamic volume provisioning.
+
+*Datastore Clusters* can only be used for the first purpose as it cannot be
+specified directly in [vSphere cloud configuration][vsphere-cloud-config].
+
+There are two places where Datastores and Datastore Clusters can be configured
+in KKP
+
+- At datacenter level (either with [Seed CRD]({{< ref "../../concepts/seeds" >}})
+    or [datacenters.yaml]({{< ref "../../concepts/seeds" >}})) is possible to
```
toschneck, line 216: Isn't the datacenter.yaml already depricated?

irozzo-1A: yes, it is but 2.15 still supports it and the majority of our customers is still using it

toschneck (commented):
@irozzo-1A  Not sure if KKP change back to Kbuermatic was on purpose?
besides of this it looks good to me

content/kubermatic/master/requirements/cloud_provider/_vsphere.en.md
toschneck: ok, if its still supported  then we need to mention it

irozzo-1A:
> @irozzo-1A Not sure if KKP change back to Kbuermatic was on purpose?
> besides of this it looks good to me

I just copied what gets generated in kubermatic repo: https://github.com/kubermatic/kubermatic/blob/master/docs/zz_generated.seed.yaml

content/kubermatic/master/requirements/cloud_provider/_vsphere.en.md
irozzo-1A: BTW, the link for `datacenters.yaml` is wrong on this line. I will fix it

toschneck:
Ok we shouldn't we automate this? I mean the chance is high that it doesn't get copied on changes

For the name changes, I think the new KKP names are the correct ones, so we may should change it in the source as well

irozzo-1A:
> Ok we shouldn't we automate this? I mean the chance is high that it doesn't get copied on changes
Yes, definitely it's better to automate it or just put the link, cheaper solution. I don't think we should do it in the scope of this PR.
> For the name changes, I think the new KKP names are the correct ones, so we may should change it in the source as well
If we want KKP we should fix it on Kubermatic repo indeed.

toschneck:
> Yes, definitely it's better to automate it or just put the link, cheaper solution. I don't think we should do it in the scope of this PR.

agree,  I just thought we already do it :sweat_smile: 
hmm maybe it's just broken I saw old commits from prow https://github.com/kubermatic/docs/commit/879633a4028913faa5a900633ea5c6e9f2fcd1b3

> If we want KKP we should fix it on Kubermatic repo indeed.

I think so, it was change by https://github.com/kubermatic/docs/pull/441 That's why I wouldn't overwrite it.

youssefazrak:
@irozzo-1A lgtm minus the KKP being replaced by Kubermatic.

irozzo-1A:
@toschneck @youssefazrak I think I address your point.

PR #462, opened 2020-10-14
updated whitelist with latest dependencies + adding flatcar

toschneck:


xrstf:
/override pre-docs-spellcheck

This test fails because KKP upstream has a typo. Even fixing it here will be undone quite quickly.

PR #477, opened 2020-11-02
Fix broken links

xrstf:
I let a linkchecker check ... well, the links, and then fixed the broken ones. And then, because the PR is already boring and long, I fixed trailing whitespace issues as well.


xmudrii (approved):
/lgtm

PR #478, opened 2020-11-03
Stay on the same page when switching between product versions

xrstf:
Previously, our code for the product version selector was very simple: Everytime someone changes the version, they are loading that version's homepage. That is bad UX.

This PR makes the version switch a bit more clever:

* If the current page exists in the other version, stay on the page.
* If the current page is not a section (i.e. it's not a `_index.en.md` file), then try the section (e.g. when on `kubermatic/v2.15/requirements/cloud_provider/_azure.md`, try `kubermatic/v2.15/requirements/cloud_provider/_index.en.md`).
* Otherwise, check the parent sections until we find one that does exist. This will stop at the latest when reaching a product's homepage.

One could call this "best effort" file matching. The result looks like this, for example:

screenshot-2020-11-03-155608

Thanks to Go's templating syntax, the code looks horribly complicated. This is the best I can offer.

And since I mutated the `menu.html` so much, I opted to also harmonize the whitespace insie the templating syntax. Sorry, I could not resist the urge.

### DIFF
--- a/layouts/partials/menu.html
+++ b/layouts/partials/menu.html
@@ -1,4 +1,4 @@
-<nav id="sidebar" class="{{if $.Site.Params.showVisitedLinks }}showVisitedLinks{{end}}">
+<nav id="sidebar" class="{{ if $.Site.Params.showVisitedLinks }}showVisitedLinks{{ end }}">
   {{ $currentNode := . }}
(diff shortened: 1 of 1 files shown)


kgroschoff (commented):
/lgtm

PR #479, opened 2020-11-03
Document custom CA's in KKP

xrstf:
It's been long overdue to document how to use custom CA's in KKP. This PR finally adds a short section detailling the two options (using cert-manager or managing everything yourself).

The PR is on hold because the configuration for external CA's in IAP is not yet actually possible.


chrkl:
Looks good so far. As an alternative approach it might also be worth to describe the option to deploy a wildcard certificate as default cert for the ingress controller.

toschneck:
Just for the reference - I created a doc in the latest customer projects, maybe there is something helpful to take over https://github.com/kubermatic/customer-projects/tree/master/components/self-signed-ca

@chrkl / @xrstf can you have a look?

xrstf:
@toschneck In https://github.com/kubermatic/customer-projects/tree/master/components/self-signed-ca#apply-config-changes-of-kubermatic you mention that the CA Bundle is base64 encoded. This is not true for the KubermaticConfiguration, this was only done in the ol' Kubermatic Helm chart. Since we have the operator, the CA Bundle is just the PEM encoded certs.

toschneck:
> @toschneck In https://github.com/kubermatic/customer-projects/tree/master/components/self-signed-ca#apply-config-changes-of-kubermatic you mention that the CA Bundle is base64 encoded. This is not true for the KubermaticConfiguration, this was only done in the ol' Kubermatic Helm chart. Since we have the operator, the CA Bundle is just the PEM encoded certs.

thanks thats true :+1:

PR #480, opened 2020-11-04
Update description of the addon form fields

maciaszczykm:
Fixes https://github.com/kubermatic/dashboard/issues/2728.


xrstf:
This is so useful, I was wondering if we want to backport this into 2.15. @maciaszczykm do you think this is easily possible?

maciaszczykm:
@xrstf I am not sure about the frontend because we were doing quite big refactoring recently. The backend should be easy to merge directly into v2.15.

xrstf (approved):
Then let's tackle a backport at a later time, if ever.

/approve

PR #483, opened 2020-12-02
Fixing 1 typo in page Expose Strategy

didier-durand:
Hi there, 
As per title.
Didier


didier-durand:
Will re-open with proper sign-off. Sorry for the noise. Didier

PR #484, opened 2020-12-02
Fixing 1 typo in page Expose Strategy

didier-durand:
Hi there,
as per title
Signed-off-by: Didier Durand <durand.didier@gmail.com>


xrstf (commented):
Love it!

It breaks the fun of a single-line change a bit, but can you please make the same fix in the content/kubermatic/v2.15, so people can see the fix now instead of when we release 2.16. :)

didier-durand:
@xrstf : Christoph, the fix in 2.15 is now done through a second commit in this PR. Shall I done same for other typos in doc that I may find? Didier

xrstf (approved):
/approve

PR #527, opened 2021-02-10
Offline mode guide

mfranczy:
Signed-off-by: Marcin Franczyk <marcin0franczyk@gmail.com>


mfranczy:
@zreigz i've changed names of some root directories as they had spaces

PR #540, opened 2021-02-16
Add cookie banner

serg:
/assign @KrissiWi

### DIFF
--- a/layouts/index.html
+++ b/layouts/index.html
@@ -10,6 +10,7 @@
   {{partialCached "css-includes" .}}
 </head>
 <body>
+  {{partial "cookie-banner.html" .}}
   {{partial "gtm-noscript.html" .}}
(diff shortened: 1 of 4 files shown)


xrstf:
@KrissiWi The documentation website is static HTML. It _does not set any cookies_. A cookie banner is completely useless, creates friction, annnoys users and is generally pointless anyway.

Let's please not put this in here just so we can say "GDPR? Check."

serg:
@xrstf please note: Google Tag Manager is using on this website. And it adding bunch of scripts (Google Analytics, etc) like we have on general domain. In this case we need to use cookie banner.

As option, we could stretch banner to full width of page and put it at bottom. The height of page will decrease (on desktop view), but users could see all sections on page. The banner hides when user gives consent for it.

PR #588, opened 2021-03-03
document hetzner improvements in KKP 2.16.3

xrstf:


kron4eg (approved):

kron4eg:
@xrstf I wanted to say that this text also should be in the master too. Otherwise we risk to forget about it.

PR #602, opened 2021-04-06
Added a page about syntax and content to MachineDeployment object.

dharapvj:
Also added information on how one can use yaml based MachineDeployment instead of terraform output based MachineDeployment.

Right now, this page link appears as first in the Guide section. I do not know how to control the link order in left nav. Please let me know right position and I can fix it.


kron4eg:
/lgtm

@xmudrii PTAL

content/kubeone/master/references/kubeone_cluster_v1beta1/_index.en.md
```diff
@@ -4,6 +4,8 @@ date = 2021-02-10T12:00:00+02:00
 weight = 1
 +++
 
+This page describes API documentation (yaml file layout) for the kubeone configuration file (default name kubeone.yaml).
```
kron4eg, line 7: Please don't edit this page, this a copy of autogenerated API doc in kubeone repository

content/kubeone/master/guides/using_machine_deployment/_index.en.md
```diff
@@ -0,0 +1,19 @@
++++
+title = "Using Machine Deployments"
+date = 2021-04-05T12:00:00+02:00
+enableToc = true
++++
```
xmudrii, line 5: This document seems to be similar to a document we already have: https://docs.kubermatic.com/kubeone/master/guides/machine_controller/

Can you extend the original document with the needed bits, so that we have only one document?

content/kubeone/master/references/kubeone_cluster_v1beta1/_index.en.md
dharapvj: OK.. Let me address this. But if you look at similar documentation from other places like [Kops API documentation](https://kops.sigs.k8s.io/cluster_spec/), the page is far more reader friendly. I believe that if we do not edit it here, we need to probably add some contextualized header in the page where it gets generated in first place. Again that's just my thought.

Consequently, all the updates that I did to the table.. Will that also need to be done elsewhere? e.g. AzureSpec, AWSSpec etc? Could you let me know where can I make upstream updates so that generated page will include these changes

content/kubeone/master/guides/using_machine_deployment/_index.en.md
dharapvj: ok let me try to merge the two

content/kubeone/master/references/kubeone_cluster_v1beta1/_index.en.md
kron4eg: As I've said, the doc is autogenerated in kubeone. It uses kubeone v1beta1 API to generate and take comments from there. So if you edit the comments in that API then yes — your changes will end up here, in autogenerated doc.

https://github.com/kubermatic/kubeone/blob/master/hack/apidoc-gen.sh here's the script to generate the doc.

content/kubeone/master/guides/using_machine_deployment/_index.en.md
dharapvj: I have merged this.. as requested. please check

content/kubeone/master/references/kubeone_cluster_v1beta1/_index.en.md
dharapvj: I have removed this page change.. as requested. Please check

kron4eg (approved):

PR #686, opened 2021-05-06
Use convert-kubeconfig command instead of referencing external script

michalvanco:
The functionality of the original [script](https://github.com/kubermatic/kubermatic-installer/blob/master/kubeconfig-serviceaccounts.sh) was incorporated in the `kubermatic-installer` so this change removes the external reference and provides a direct usage example of the command.

(based on the info from @xrstf, the script won't be maintained and updated anymore).

Update is done in both CE/EE guides and in master + 2.17 version.

There is a related update of the original script (to make it work with latest yq) - kubermatic/kubermatic-installer/pull/136


xrstf (approved):
❤️ 

/approve

PR #710, opened 2021-06-07
How to cause a machinedeployment rolling restart

kron4eg:


WeirdMachine:
Wouldn't this fit better to the Cheat Sheets section?

kron4eg:
@WeirdMachine good point! Changed.

PR #714, opened 2021-06-08
Updating 404 error links and images in KKP Master and V2.17

fairhaven:


content/kubermatic/master/guides/addons/kubeflow/_index.en.md
```diff
@@ -102,7 +102,7 @@ it is possible to point the Kubeflow addon to this KKP OIDC service. In case tha
 
 This setup however requires some configuration on the KKP platform side as well. The KKP installation administrator
 has to add the following section into the KKP’s Helm `values.yaml` before installing the `oauth` chart
-(see the [Securing System Services documentation](../../../installation/securing_services) for more details):
+(see the [Securing System Services documentation](../../../guides/kkp_security/securing_system_services/) for more details):
```
kron4eg, line 105: I'd like to suggest to avoid using plain markdown links for internal cross-linking. If instead we'd use hugo macros ref, it will signal us that we are making some pages 404 in case of changes (means, hugo will blow up on build stage).

See more at: https://gohugo.io/content-management/cross-references/#use-ref-and-relref

fairhaven: You mean using for example: ({{ref: ../../.......}}) instead (../../......) for the links?

PR #722, opened 2021-06-24
Merging Global Settings and Admin Panel content 

fairhaven:


content/kubermatic/master/tutorials_howtos/administration/admin_panel/_index.en.md
```diff
@@ -4,42 +4,106 @@ date = 2018-08-09T12:07:15+02:00
 weight = 40
 +++
 
-The Admin Panel is a place for the Kubermatic administrators where they can manage the global settings that directly
-impact all Kubermatic users. Admin rights can be granted from the admin panel itself and also from the kubectl by
+The Admin Panel is a section where the Kubermatic administrators can manage the global settings that
```
maciaszczykm, line 7: Why changing "place" to "section"?

content/kubermatic/master/tutorials_howtos/administration/admin_panel/_index.en.md
```diff
@@ -4,42 +4,106 @@ date = 2018-08-09T12:07:15+02:00
...
-After logging in to the dashboard with an administrator you should be able to access the admin panel from the menu up
+After logging in to the dashboard as an administrator, you should be able to access the admin panel from the menu up
 top.
 
 ![](/img/kubermatic/master/ui/admin_panel_access.png?height=300px&classes=shadow,border "Accessing the Admin Panel")
 
 ![](/img/kubermatic/master/ui/panel.png?height=350px&classes=shadow,border "Admin Panel")
 
-The Admin Panel offers a wide set of features including but not limited to:
+## Global Settings
```
maciaszczykm, line 23: If you want to keep it as a header then perhaps let's move it to the bottom as the whole page is named admin panel, so it should have the priority. We can also remove the header and mention the CRD existence with something like: 

```
Global settings can be also modified from the command line with kubectl. It can be done by editing the globalsettings KubermaticSetting CRD. This resource has following structure...
```

To be honest I'd prefer second option.

content/kubermatic/master/tutorials_howtos/administration/admin_panel/_index.en.md
```diff
@@ -4,42 +4,106 @@ date = 2018-08-09T12:07:15+02:00
...
+It can be edited directly from the command line:
+
+```
+$ kubectl edit kubermaticsetting globalsettings
+```
+
+**Note:** Custom link `icon` is not required and defaults will be used if field is not specified. `icon` URL can
+point to the images inside the container as well, i.e. `/assets/images/icons/custom/github.svg`.
+
+## Manage Global Settings Via UI
```
maciaszczykm, line 85: This would not be needed if we reorganize the order/remove the global settings section.

content/kubermatic/master/tutorials_howtos/administration/admin_panel/_index.en.md
fairhaven: Because a place looks awkward in the context since it is a doc, not a location or physical structure. However, I am thinking of using it like this:  "The Admin Panel is "where" the Kubermatic administrators................"

maciaszczykm: Yes, but this is not related to the doc but to the place/view in the app.

fairhaven: Alright. I can change it if it suits better

PR #732, opened 2021-07-27
Add instructions to create the first KKP admin after installation

lsviben:
Adds instructions to create the first KKP admin after installing


lsviben:
@shaase-ctrl this is hanging from some time. PTAL

PR #742, opened 2021-08-05
Add User Cluster MLA Docs

rastislavs:
Signed-off-by: Rastislav Szabo <rastislav@kubermatic.com>


jiachengxu (commented):
PR really looks good to me!
I just have one small comment: do you think it makes sense to have a table of content at the beginning of every .md file? For example: https://docs.kubermatic.com/kubermatic/v2.17/guides/addons/kubeflow/
This can be achieved by adding `{{< table_of_contents >}}` to the beginning of the file.
But this is not a blocker for the PR, we can add that later

content/kubermatic/master/guides/monitoring_logging_alerting/user_cluster/admin_guide/_index.en.md
```diff
@@ -0,0 +1,301 @@
...
+
+```bash
+kubectl get pods -n mla
+```
+
+- If there are any pods crashing, review their logs for the root cause
+
+## Uninstallation
+
+**WARNING:** Before proceeding with any of the following steps, make sure that you backup all data that you may still need - metrics data / logs in the object store, alertmanager / rules configuration, Grafana dashboards.
```
jiachengxu, line 264:
```suggestion
{{% notice warning %}} 
Before proceeding with any of the following steps, make sure that you backup all data that you may still need - metrics data / logs in the object store, alertmanager / rules configuration, Grafana dashboards.
{{% /notice %}}
```
This will look like the warning block at the top of this doc: https://docs.kubermatic.com/kubermatic/v2.17/guides/addons/kubeflow/

rastislavs:
> do you think it makes sense to have a table of content at the beginning of every .md file? For example: https://docs.kubermatic.com/kubermatic/v2.17/guides/addons/kubeflow/
> This can be achieved by adding `{{< table_of_contents >}}` to the beginning of the file.

I have used `enableToc = true` on some places where I thought it is needed, but maybe it would be worth to add it to some more places as well

jiachengxu:
> I have used enableToc = true on some places where I thought it is needed, but maybe it would be worth to add it to some more places as well

Oh, I didn't notice that you added that to user guide and admin guide, then I think it is already good enough

PR #755, opened 2021-08-16
fix featureGates format in etcd-launcher docs

judge-red:
The previously used format did not appear to work in the latest KKP version.
Supposedly the format changed at some point without this doc page being
updated to reflect that change.

Signed-off-by: Sandro Mathys <sandro@mathys.io>


WeirdMachine:
Fixed by https://github.com/kubermatic/docs/pull/805

PR #806, opened 2021-09-16
drop flatcar support

WeirdMachine:
Remove flatcar as supported OS


WeirdMachine:
Still supported my bad

PR #840, opened 2021-10-29
Fix links to requirements and upgrade notes in CE installation

embik:
These links are correct in the EE installation document, but are not working in the CE version of the document.


xrstf (approved):
/approve

Issue #839, opened 2021-10-27
Sidebar Design Update


cschieder:
### Design
![image](https://user-images.githubusercontent.com/49145827/139012374-ac400c2d-e324-4f34-8726-123520446884.png)

### To do
- [x] Remove Product Logo (On all products)
- [x] Make every sublevel hidden by default and add an arrow to unfold (if no subpoints - then no arrow)
- [x] If an area is unfolded the arrow turns to point up instead of down
- [x] Hover left, on click right top 30px, bottom 25px (30px to be used for the first level navigation, 25px for the navigation points below)
![image](https://user-images.githubusercontent.com/49145827/139011742-c1a0415b-db01-4f02-8437-916fa6ae63d9.png)

- [x] Area of response (if possible I want the area where the unfold button reacts to be this size, the colors are just to make it visible 50px blue, 50px by 30px green)
![image](https://user-images.githubusercontent.com/49145827/139011907-6036f791-cfa7-420b-8594-d5930d1b332c.png)

PR #842, opened 2021-11-02
move sections for KKP master and update sidebar menu behaviour

serg:


serg:
fix #839

xmudrii:
@serg I have two concerns about this PR:

* We also need to update `master` docs to the new structure. Otherwise, 2.19 docs will use the old structure because we are always copying from the master docs when cutting a new minor release
  * Note that you can't just copy 2.18 docs to master docs, you'll need to do the reorganization manually. That's because the master docs mostly like already have 2.19 content
* Touching the 2.18 docs means that the old URLs will not work (please correct me if I'm wrong). Since 2.18 is already out for a few months, folks might have references to docs, and after this PR, all those references/links will lead to 404
  * I'd prefer that we only touch master docs, so the new structure is only in effect for 2.19. This means 2.18 will use the old structure, but that sounds safer
  * Redirects might also be possible (and we did that in the past for KubeOne), but it can get pretty complicated when you have a lot of docs, so it's probably better to avoid them

cc @shaase-ctrl

serg:
@xmudrii, your points sounds reasonable.
I thought we need to move sections for current version of docs (v2.18). I didn't find strictly requirements, that this must be only for `master` version.

Therefore, we can revert content changes for v2.18 and move the sections only for `master` branch. Therefore v2.18 will be displaying in previous state. @shaase-ctrl please confirm those actions.

serg:
@xmudrii, the changes were reverted for v2.18 and applied for `master` version. Please check and approve.

xmudrii (approved):

PR #843, opened 2021-11-03
Remove traces of kubermatic-operator-util from docs of versions after removal

dermorz:
Reference: https://github.com/kubermatic/kubermatic/issues/6946


xrstf (approved):
/approve

PR #854, opened 2021-11-17
Remove helm2 references

vgramer:
Since KKP V2.18 we do not support helm 2 anymore. This PR remove helm 2 references from the documentation.


xrstf (approved):
/approve

PR #859, opened 2021-11-19
Add admissionPlugins documentation for custom admission plugins

embik:
We introduced a field `spec.admissionPlugins` quite a while ago that offers a flexible way to append admission plugins, but we never surfaced that in the documentation. For easy resolution of https://github.com/kubermatic/edgematic/issues/52, we can document this and thus, help users that have custom needs for their admission plugin list.

Fixes: https://github.com/kubermatic/edgematic/issues/52

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,6 @@ tutorials-tests/cypress/fixtures
 
 # IDE specific files
 .idea/
+
+# Hugo
+.hugo_build.lock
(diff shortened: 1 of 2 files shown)


embik:
@zreigz, you added this field quite a while ago in https://github.com/kubermatic/kubermatic/pull/5044, so I'd wanted to ask you for review and hear your thoughts on exposing this in the documentation.

PR #876, opened 2021-12-09
Use https for kubeone.io

simon-wessel:
The install shell script for kubeone should be fetched using https.
Made possible by https://github.com/kubermatic/website/pull/1767


kron4eg (approved):
/approve

PR #889, opened 2022-01-05
Use empty map for datacenters example in Seed installation

embik:
Our docs give `datacenters: []` in the example `Seed` resource. That field is a map at least since v2.16 (https://github.com/kubermatic/kubermatic/blob/release/v2.16/pkg/crd/kubermatic/v1/datacenter.go#L80) and the example YAML fails webhook validation as it's invalid.


kron4eg (approved):
/approve

PR #904, opened 2022-01-13
Add information about changed ingress annotations.

wurbanski:
Updated nginx ingress uses `nginx.ingress.kubernetes.io` annotations.


xrstf (approved):
/approve

PR #912, opened 2022-01-18
updated OPA docs, removing rawJSON references

lsviben:
updated OPA docs, mostly removing rawJSON references


imharshita (approved):
/lgtm

PR #915, opened 2022-01-18
Improve search behaviour

serg:


themes/hugo-theme-learn/static/js/search.js
```diff
@@ -1,89 +1,128 @@
...
+  var getSearchOptions = function(searchResultsBox) {
+    var searchFirstRun = true;
 
-function endsWith(str, suffix) {
-  return str.indexOf(suffix, str.length - suffix.length) !== -1;
-}
+    var options = {
+      searchClient: algoliasearch(
+        'Q03DAQELCN',
+        'f8b7a0497068023efe046671dd40ec62'
```
xmudrii, line 8: What is this? If this is an API key, should we really put it publicly in the source code?

serg: This is public key for using search API

xmudrii (approved):
/lgtm
/approve

PR #927, opened 2022-01-26
Add deprecation notice to dynamic kubelet configuration

embik:
Dynamic kubelet configuration is deprecated upstream. This adds a note to the documentation and removes the note saying the feature is in beta.

I'm also adding it to 2.19 documentation so users are aware of it and this isn't hidden away in `master` only.


ahmedwaleedmalik (approved):
/lgtm
/approve

PR #934, opened 2022-02-01
Simplify docs creation for releases

### LINKED ISSUE #929: Reduce manual effort when preparing docs for a release
Reduce/eliminate manual effort for:
- Copy content from `/master` to `/vN.NN`
- Update all references within `/vN.NN`
- Update `data/products.yaml` to add new release in the right place
- Update supported versions of KKP components

dermorz:
Closes #929 

Any remarks are welcome. :)

### DIFF
--- a/hack/prepare-release.sh
+++ b/hack/prepare-release.sh
@@ -76,7 +76,7 @@ KUBERMATIC_DIR=${KUBERMATIC_DIR:-"../kubermatic"}
 if [[ $PRODUCT == 'kubermatic' ]]
 then
   tmpfile=$(mktemp)
-  head -13 content/kubermatic/master/architecture/support_policy/KKP_components_versioning/_index.en.md > $tmpfile
+  sed '/^|.*KKP/Q' content/kubermatic/master/architecture/support_policy/KKP_components_versioning/_index.en.md > $tmpfile
   version_table >> $tmpfile
   mv $tmpfile content/kubermatic/master/architecture/support_policy/KKP_components_versioning/_index.en.md
 
--- a/hack/prepare-release.sh
+++ b/hack/prepare-release.sh
@@ -14,10 +14,11 @@ usage() {
   echo "Flags:"
   echo "  -p    Product selection. One of 'kubermatic', 'kubeone', 'kubecarrier'. (env: PRODUCT)"
   echo "  -v    Version of the upcoming release. (env: VERSION)"
+  echo "  -k    Location of kubermatic/kubermatic working copy. (env: KUBERMATIC_DIR, default: '../kubermatic')"
   echo "  -h    Print this help."
   echo
   echo "Product and Version can be passed by flag or environment variable. The flag has the higher weight."
-  echo "For the component version update the code of kkp needs to be located at ../kubermatic."
+  echo "For the component version update the code of KKP either needs to be located at ../kubermatic or an alternative location needs to be set (see -k)."
 }
 
 line() {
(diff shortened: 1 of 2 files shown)


hack/prepare-release.sh
```diff
@@ -0,0 +1,95 @@
...
+  (usage; exit 1)
+
+[[ $PRODUCT =~ ^(kubermatic|kubeone|kubecarrier)$ ]] ||
+  (usage; exit 1)
+
+# Update component versions and copy static images only when preparing docs for KKP release
+if [[ $PRODUCT == 'kubermatic' ]]
+then
+  tmpfile=$(mktemp)
+  head -13 content/kubermatic/master/architecture/support_policy/KKP_components_versioning/_index.en.md > $tmpfile
```
SimonTheLeg, line 72: I think this is quite brittle for two reasons:
1. It copies over the date timestamp from the previous document. This results in all future documents having the same timestamp
2. It is highly dependent that the content stays within 13 lines. So if someone were to add a line in the file, it would not work correctly anymore. Additionally the script would not break, but just produce a wrong document. This is quite dangerous as someone could easily miss it, if that person does not manually check the file every time. 

To solve this I could think of three possible solutions:

### A) Use sed to anchor on a piece of text

Either this could be binding to the table on the "| KKP Components" somehow. Or you could add a line "compatibility matrix" above the table and bind to this. Anyways the idea is basically to anchor to a piece of text and copy anything above it. The advantage is that this either works when the number of lines change or will fail if the anchoring text has changed.

And then to solve the date timestamp, you can just use a shell function to insert it.


### B) Create a gohugo archetype

In this solution you would add a "Components Versioning" archetype, and then let gohugo create your new file. Basically the archetype would contain the 12 lines. And then you simply create the new file using the `hugo new` command.

In this case, the timestamp would automatically be created by hugo.

This solution of course then has hugo as a dependency for the script.


### C) Hybrid between the two

If you don't want to use hugo, I think both ideas could also be combined. Basically create a markdown file with the 12 lines to be used as a template. And then just use a bash utility to generate the timestamp.




If you choose B) or C), please also add a markdown commend in the final doc that indicates that it is auto-generated.

SimonTheLeg (changes requested):

hack/prepare-release.sh
dermorz: Good points, I agree this hardcoded approach will break if the upper part of the page changes. I will change it to a more dynamic approach.

---

Regarding the document creation timestamp: This holds true for all the content that is copied over. If I change this here, I actually have to set it on every single markdown document I am copying over from `master` to `<version>`. As far as I have checked docs of previous releases, this has not been done before. I can introduce it everywhere, but is it really worth it? It would also only affect the metadata of the docs for new releases.

hack/prepare-release.sh
```diff
@@ -0,0 +1,95 @@
...
+}
+
+line() {
+  printf "| %-30s | %-30s |\n" "$1" "$2"
+}
+
+version_table() {
+  line "KKP Component" "Version"
+  line "---" "---"
+  for comp in $(find ../kubermatic/charts -name 'Chart.yaml' -printf '%h\n')
```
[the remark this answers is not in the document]
dermorz: Good hint about MacOS `find`. I will change it to trim the suffix (filename) as I did with the prefix (relative path) using bashparameter expansions and temp vars.

I also thought about adding a parameter to optionally set the location of a `kubermatic/kubermatic` working copy using the current value as default. Your comment emphasises that I should have done that, so I will add it. :+1:

hack/prepare-release.sh
SimonTheLeg: ah I see. I did not know that the final timestamp does not show up in the document. Then it would be fine from my side to leave it as is

SimonTheLeg:
/hold 
do not merge the PR right away after first approval

SimonTheLeg (approved):

xmudrii (approved):
/lgtm
/approve

PR #960, opened 2022-02-09
fix link inside tab

SimonTheLeg:
This was the only possible way I got this to work, as it seems you
cannot use Markdowns Reference-style Links with the tab shortcode.


ahmedwaleedmalik (approved):
/approve

PR #974, opened 2022-02-15
kubeone: add a note about terraform refresh

xmudrii:
Relevant to https://github.com/kubermatic/kubeone/issues/1847

/assign @vgramer @kron4eg @ahmedwaleedmalik


ahmedwaleedmalik (approved):
/approve
/lgtm

vgramer (approved):
/lgtm

kron4eg:
There is special waiting built-in azure terraform config

https://github.com/kubermatic/kubeone/blob/c46b9fb051afe0e3db91f92409a895838505c4b9/examples/terraform/azure/main.tf#L276-L283

xmudrii:
@kron4eg TIL, but looks like it's not enough. We can eventually increase it to 1 minute. However, I would still leave this warning because such errors are not easy to fix, so it's better to double check before provisioning the cluster.

PR #978, opened 2022-02-17
Update instructions for accessing k8s cluster via start.kkp

michalvanco:
This steps depends on used Terraform backend, for github - s3/gcs backend it used,
with GitLab - terraform state on GitLab is used and couple of variables are required.


ahmedwaleedmalik (approved):
/approve
/lgtm

PR #984, opened 2022-02-21
Updated the download repository name as per recent renaming changes

archups:
As per discussion and suggestion from Christoph Mewes, the repository `kkp-on-node` is renamed to `kkp-single-node`. Hence PR to update the clone repository link here. 

Signed-off-by: archups <archupsawant@gmail.com>


kron4eg (approved):
/approve

PR #985, opened 2022-02-22
Update CRD docs, fix rendering issues

wurbanski:
Due to the depth of type hierarchy, some fields were not rendered properly. This fixes the CRD reference and the config file for automation.


kron4eg (approved):
/approve

PR #991, opened 2022-03-03
Add documentation for Nutanix support

### LINKED ISSUE #990: Add documentation for Nutanix cloud provider
Our Nutanix support is not well documented in the docs at the moment (not at all, to be honest), so we should start adding it to our documentation.

https://docs.kubermatic.com/kubermatic/master/architecture/support_policy/provider_support_matrix/ here, for example.

embik:
Fixes #990.

### DIFF
--- a/content/kubermatic/master/architecture/support_policy/provider_support_matrix/_index.en.md
+++ b/content/kubermatic/master/architecture/support_policy/provider_support_matrix/_index.en.md
@@ -21,6 +21,7 @@ Kubermatic Kubernetes Platform supports a multitude of different cloud providers
 * OTC (Open Telekom Cloud)
 * Bare metal via KubeVirt
 * Hetzner
+* Nutanix (requires Prism Central)
 
 **Tier 3**
 * Alibaba Cloud
--- a/content/kubermatic/master/architecture/support_policy/provider_support_matrix/_index.en.md
+++ b/content/kubermatic/master/architecture/support_policy/provider_support_matrix/_index.en.md
@@ -16,12 +16,12 @@ Kubermatic Kubernetes Platform supports a multitude of different cloud providers
 * vSphere beginning with v6.5
 * OpenStack (Releases with maintenance or extended maintenance)
 * KubeVirt
+* Nutanix (requires Prism Central)
 
 **Tier 2**
 * OTC (Open Telekom Cloud)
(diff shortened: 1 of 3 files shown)


embik:
@shaase-ctrl can you please take a look if the wording here is consistent with how we want to present our Nutanix support? Is "Tier 2" the correct tier for it?

shaase-ctrl:
It is Tier 1 support. We have consistent test setup etc.

embik:
Thanks, updated.

maciaszczykm:
It would be nice to have somewhere default values for Nutanix configuration like port, storage container etc., so we could link to the docs in the UI.

embik:
cc @WeirdMachine because he might better know about some of the CSI side stuff.

maciaszczykm:
>but storage containers and everything else cannot really be defaulted since it depends on the user's Nutanix setup


It seems to be defaulted looking at the comments: https://github.com/kubermatic/kubermatic/blob/57473080aa76f0e07726f137ea395ac98d8b00b1/pkg/apis/kubermatic/v1/cluster.go#L845

I cannot pre-fill the values in the UI as this will block the presets and you would need to remove the default values to make preset selector enabled. So I would like to add hints and point to the docs where the information about default values can be found. We may of course skip it but then the user will not know what happens if the fields will be left empty.

shaase-ctrl:
But do we document that we set these standards?

maciaszczykm:
If not, then I can put the defaults in the UI or not do it at all. What would you prefer @shaase-ctrl?

shaase-ctrl:
I would love to have both. Standards in documentation -> content worth googling. Displayed in UI -> Usability.

embik:
@shaase-ctrl @maciaszczykm please take another look, I've added the defaults for the CSI driver config.

maciaszczykm:
LGTM

WeirdMachine:
I think we should clarify the connection requirements for Prism Central and Prism Element. Prism Central needs to be accessed by the seed-cm so from the seed cluster and Prism Element by the User Cluster.

embik:
> I think we should clarify the connection requirements for Prism Central and Prism Element. Prism Central needs to be accessed by the seed-cm so from the seed cluster and Prism Element by the User Cluster.

Fair point, let me update that.

PR #1008, opened 2022-03-16
Add a note to EtcdLauncher feature flag

embik:
We don't recommend enabling the feature flag at the same time as an upgrade. This adds a note.


xrstf (approved):
/approve

PR #1012, opened 2022-03-17
Remove CRD reference from v2.19 docs, CRDs are not available there

embik:
We have CRD reference docs for the `kubermatic.k8c.io/v1` API group in our docs for KKP v2.19 for whatever reason. But that API group isn't available to users in v2.19, it just existed in the code internally to prepare the migration.

@wurbanski @xrstf (you worked/approved these things I believe) let me know if I missed something here, but I think this should be removed.


ahmedwaleedmalik (approved):
/approve

/hold for @xrstf to take a look

PR #1021, opened 2022-03-22
Restore KKP monitoring runbook

### LINKED ISSUE #1017: Restore alerting runbook
We used to have https://docs.kubermatic.com/kubermatic/master/monitoring/runbook/ but no more, which is a mistake and the runbook should be restored.

kron4eg:
Fixes #1017

### DIFF
--- a/content/kubermatic/master/monitoring/_index.en.md
+++ b/content/kubermatic/master/monitoring/_index.en.md
@@ -1,0 +1,10 @@
++++
+title = "Monitoring"
+date = 2018-04-28T12:07:15+02:00
+weight = 23
+chapter = true
++++
+
+# Monitoring
+
+This chapter describes the integrated monitoring stack and how to customize it to your needs.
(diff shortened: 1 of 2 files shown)


wurbanski:
Under the new docs structure, the runbooks should fall under cheat_sheets/alerting_runbook/ - actually there is something there, just likely not being automatically updated. I see 2 options:
1. we can change the link in the laerts 
2. set up a redirect from old address to new (is it possible in Hugo?) and make sure the content is being generated in correct place

WDYT?

kron4eg:
@wurbanski the links to the master/monitoring/runbook are external links (generated in charts/monitoring/prometheus/rules in kubermatic). I'm not sure we can change old releases to make them point to some new location easily.

@xrstf any opinions?

wurbanski:
We could change the URL in new releases and still alias the new location to be accessible from the old one like here: https://gohugo.io/content-management/urls/#aliases

PR #1022, opened 2022-03-22
Use search by product and version

### LINKED ISSUE #981: Search returns results from different versions
It looks like the search for docs.kubermatic.com has been updated recently (not sure when). Since then, searching for a key word gives you results from seemingly random KKP versions, not just the one I've selected. 

It's very hard to find the document I'm looking for in the right version.

Example:

Screenshot 2022-02-18 at 14 10 21

I would expect that the search prefers the version I have selected and maybe even omits the `master` version from searches if a stable release is selected

serg:
fix #981 
/assign @embik


embik (approved):
This is awesome, thanks for fixing it. Works fine from the preview deployment.

/approve

serg:
You are welcome, glad to help!

PR #1025, opened 2022-03-24
Update examples for KKP v2.20

dakraus:
Updated some examples to include the changes to the CRDs introduced in KKP v2.20.0:
* Change `KubermaticConfiguration.spec.featureGates` to `object (keys:string, values:boolean)`
* Use `requiredEmails` instead of `requiredEmailDomains`
* Use `userClusterMLAEnabled` instead of `user_cluster_mla_enabled`


embik:
@dakraus thanks for fixing those! Minor request, could you also mirror your changes to the `master` directory? All these changes remain true for KKP 2.21 and beyond, so let's update it for future documentation as well.

embik (approved):
/approve
/lgtm

Thanks @dakraus!

PR #1076, opened 2022-06-02
Use title case consistently across all pages

embik:
The king of all nitpick PRs, this one applies [title case](https://en.wikipedia.org/wiki/Title_case) across all documentation pages. Most of the page titles were already following this, but a few of them didn't. I fixed them because they were my ultimate pet peeve.

I've done this for `master` and 2.20, since that's the documentation people get to see by default when going to docs.kubermatic.com.


ahmedwaleedmalik (approved):
/approve

PR #1098, opened 2022-07-07
Clarify KubeVirt status that it's tech-preview

mfranczy:
We only have this info for the machine controller, we have to update it everywhere that's visible for everyone.


mfranczy:
@shaase-ctrl please take a look

PR #1107, opened 2022-07-21
Add documentation on how to customize audit-logs sidecar 

embik:
We added `spec.auditLogging.sidecar` to the ClusterSpec, which allows customizing fluentbit configuration. This can be mainly used to ship audit logs to an external system instead of writing it to standard output, something that might not be desired in regulated environments.


ahmedwaleedmalik (approved):
/approve

PR #1118, opened 2022-07-28
Update CE installation reference

almereyda:
This link pointed towards itself, and now targets a location that supposedly fits the context.

Signed-off-by: jon r <jon@allmende.io>

In redoing a KubeOne + Kubermatic installation one and a half years later, I have found this dead link, which was still working the last time.

Is it correct that this is the intended target of the link?

It may be that this link is also broken in the other installation instructions, which I didn't check.


embik:
Thanks for fixing this!

/ok-to-test

embik (approved):
/approve

PR #1122, opened 2022-08-01
Add note about identifying audit logging source

embik:
By default we don't add the cluster ID to audit logs. This adds a documentation section for it. I'll probably add this by default in later KKP releases but it didn't make the cut for 2.21.


ahmedwaleedmalik (approved):
/approve

PR #1132, opened 2022-08-05
Exclude the pages with redirect from sitemap

serg:


ahmedwaleedmalik (approved):
/approve

TIL I didn't know that Hugo supported sitemap exclusions in this way. This is super nice.

serg:
@ahmedwaleedmalik Thank you. for approving!
We have custom sitemap generation. And with this field we can exclude some pages from writing to sitemap file.

PR #1137, opened 2022-08-05
kubeone local docs

kron4eg:
Documentation for the https://github.com/kubermatic/kubeone/pull/2125

### DIFF
--- a/content/kubeone/master/guides/local/_index.en.md
+++ b/content/kubeone/master/guides/local/_index.en.md
@@ -1,0 +1,68 @@
++++
+title = "One Node Cluster"
+date = 2022-08-05T15:58:00+02:00
+enableToc = true
++++
+
+## Introduction
+
+`kubeone local` subcommand helps initializing a **local** (i.e. on a local
+machine), cluster. Container runtime, control-plane and kubelet with some basics
+like CNI will be installed, together with removed taints to unblock the single
+Node for a regular workloads. We call this "all-in-one cluster".
+
+{{% notice warning %}}
+`kubeone local` will do changes to your operating system, don't run it on a
+working machine.
+{{% /notice %}}
+
+## Use Cases
+
+A all-in-one setup could be useful for cases such as:
+* Single edge server
+* Developer VM
+* CI VM
+
+## Prerequisite
+
+Local user that run the `kubeone local` ether has to have passwordless `sudo` or
+be a `root` user.
+
+### How to configure passwordless sudo
+
+Edit the `/etc/sudoers` file with:
+```shell
+sudo visudo
+```
+
+Add a line:
+```
+<your_username> ALL=(ALL:ALL) NOPASSWD:ALL
+```
+Replacing `<your_username>` with your login (i.e. the value of shell `$USER`).
+
+## Usage
+
+For `kubeone local` the manifest is **optional**! You only need a config, if you
+really want to customize something. By default manifest will be generated in
+memory.
+
+The Kubernetes api endpoint would be autodetect to the default gateway
+interface. It can also be specified using `--apiendpoint` flag.
+
+## Configuration
+
+You can provide **optional** KubeOne Cluster manifest. Some parts of it will be
+forcefully rewritten. Such fields include:
+
+* `name`: will always be `local`
+* `controlPlane`: will always be equal to local node, with Public and Private IP
+  set to detected default IP, with empty `Taints`.
+* `cloudProvider` will always be set to `None`
+* `machineController.deploy` will always be set to `false`
+* `operatingSystemManager.deploy` will be always set to `false`
+* `versions.kubernetes` default to `1.24.2`, but can be set using flag
+  `--kubernetes-version`
+
+Rest of configuration is up to you. Config will be defaulted as usual (i.e. you
+can omit specifying CNI or containerRuntime). 


content/kubeone/master/guides/local/_index.en.md
```diff
@@ -0,0 +1,62 @@
++++
+title = "Local subcommand"
```
xmudrii, line 2: Let's find a better title for this document. What about something like Creating a local all-in-one cluster?

content/kubeone/master/guides/local/_index.en.md
```diff
@@ -0,0 +1,62 @@
...
+date = 2022-08-05T15:58:00+02:00
+enableToc = true
++++
+
+## Introduction
+
+`kubeone local` subcommands helps you to initialize a **local** (e.g. on the
+local machine, in the current shell), 1 node all-in-one cluster.
```
xmudrii, line 10: It would be nice to clarify what exactly all-in-one means.

content/kubeone/master/guides/local/_index.en.md
```diff
@@ -0,0 +1,62 @@
...
+
+## Use Cases
+
+Such a setup could be useful for cases such as:
+* Single edge server
+* Developer VM (kinda competing with kind but without placing everything in
```
xmudrii, line 16: I would add a warning about this use case that this is possible, but that you have to be very careful because KubeOne modifies various system settings.

content/kubeone/master/guides/local/_index.en.md
```diff
@@ -0,0 +1,62 @@
...
+Edit the `/etc/sudoers` file with:
+```shell
+sudo visudo
+```
+
+Add a line:
+```
+<your_username> ALL=(ALL:ALL) NOPASSWD:ALL
+```
+Replacing `<your_username>` with your login (i.e. `$USER`).
```
xmudrii, line 36: I guess you can't put just `$USER`, so it's probably more correct to say the value of `$USER`.

content/kubeone/master/guides/local/_index.en.md
```diff
@@ -0,0 +1,62 @@
...
+You can provide **optional** KubeOne Cluster manifest. Some parts of it will be
+forcefully rewritten. Such fields include:
+
+* `name`: will always be `local`
+* `controlPlane`: will always be equal to local node, with Public and Private IP
+  set to detected default IP, with empty `Taints`.
+* `cloudProvider` will always be set to `None`
+* `machineController.deploy` will always be set to `false`
+* `operatingSystemManager.deploy` will be always set to `false`
+* `versions.kubernetes` default to `1.24.2`, but can be set using flag
```
xmudrii, line 58: I'm wondering if we can use version markers instead.

content/kubeone/master/guides/local/_index.en.md
kron4eg: That title will be shown is sidebar menu, and such a long like would word-wrap around and make it ugly AF.

content/kubeone/master/guides/local/_index.en.md
kron4eg: What else besides "1 node all-in-one cluster" can it mean?!

content/kubeone/master/guides/local/_index.en.md
xmudrii: What about `Local installation` instead?

content/kubeone/master/guides/local/_index.en.md
xmudrii: I thought about clarifying that it's both a control plane and a worker node. Something like:

> 1 node all-in-one cluster (the node is running both control plane and workloads)

content/kubeone/master/guides/local/_index.en.md
kron4eg: Without context it looks like it's about "installing kubeone binary locally", a bit misleading.

Few options from the top of my head:
* `One Node Cluster`
* `All-in-one Cluster`
* `Edge Node`
* `Edge Cluster`

content/kubeone/master/guides/local/_index.en.md
kron4eg: Fixed

content/kubeone/master/guides/local/_index.en.md
kron4eg: done

content/kubeone/master/guides/local/_index.en.md
kron4eg: done

content/kubeone/master/guides/local/_index.en.md
```diff
@@ -0,0 +1,64 @@
...
+`kubeone local` subcommand helps initializing a **local** (i.e. on a local
+machine), cluster. Container runtime, control-plane and kubelet with some basics
+like CNI will be installed, together with removed taints to unblock the single
+Node for a regular workloads. We call this "all-in-one cluster".
+
+## Use Cases
+
+A all-in-one setup could be useful for cases such as:
+* Single edge server
+* Developer VM (WARNING: it will do changes to your operating system, don't run
```
xmudrii, line 18: Nit: we have a warning block, we should probably use it instead.

content/kubeone/master/guides/local/_index.en.md
xmudrii: Maybe `All-in-one Cluster` is a good solution because we already use that wording in the document.

xmudrii (approved):
/lgtm
/approve

PR #1140, opened 2022-08-09
Fix broken links for internal pages

serg:
/assign @xmudrii

### DIFF
--- a/content/kubermatic/master/installation/install-kkp-on-node/_index.en.md
+++ b/content/kubermatic/master/installation/install-kkp-on-node/_index.en.md
@@ -14,7 +14,7 @@ This chapter will guide you through the KKP Master setup on a single master/work
 
 In this **Get Started with KKP** guide, we will be using AWS Cloud as our underlying infrastructure and KKP release v2.18.4.
 
-> For more information on the kubeone configurations for different environment, checkout the [Creating the kubernetes Cluster using Kubeone](https://docs.kubermatic.com/kubeone/master/tutorials/creating_clusters/) documentation.
+> For more information on the kubeone configurations for different environment, checkout the [Creating the kubernetes Cluster using Kubeone]({{< relref "../../../../kubeone/master/tutorials/creating-clusters/" >}}) documentation.
 
 ## Prerequisites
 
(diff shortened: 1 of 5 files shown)


xmudrii (approved):
/lgtm
/approve

content/kubermatic/master/installation/install-kkp-on-node/_index.en.md
```diff
@@ -14,7 +14,7 @@ This chapter will guide you through the KKP Master setup on a single master/work
 
 In this **Get Started with KKP** guide, we will be using AWS Cloud as our underlying infrastructure and KKP release v2.18.4.
 
-> For more information on the kubeone configurations for different environment, checkout the [Creating the kubernetes Cluster using Kubeone](https://docs.kubermatic.com/kubeone/master/tutorials/creating_clusters/) documentation.
+> For more information on the kubeone configurations for different environment, checkout the [Creating the kubernetes Cluster using Kubeone]({{< relref "../../../../kubeone/master/tutorials/creating-clusters/" >}}) documentation.
```
xmudrii, line 17: Do we want to link to master docs or the latest stable docs? Master docs might get changed in development and might not reflect the latest available KubeOne release.

xmudrii: In other words, do we have an easy way to link to the latest stable docs without putting the version in the URL?

serg: We don't have automatically generation for link on stable version in this case. Besides, here is possibility, that someday the link will changed. Here we need to define direct link to version, that we might want to use. This file under `master` version of product as a link where it pointing on. Therefore I haven't idea where link should point on KubeOne `v1.4`(on Kubermatic `master` page) or on other location. Could you please advice which link we need to use in this file?

xmudrii: Let's leave it as it is for now, we can always get back to this later if needed.

xmudrii (approved):

content/kubermatic/master/installation/install-kkp-on-node/_index.en.md
serg: Agrees, besides the build is fail if somebody will change page url, that doesn't exist on this page, thanks to using `relref`. Thank you for approval!

PR #1185, opened 2022-08-26
Exclude references/crds pages of KKP from search index

serg:
The reason is exceeded limit of page content size for record in search index.
/assign @ahmedwaleedmalik


ahmedwaleedmalik (approved):
/approve

PR #1200, opened 2022-08-31
Fix some broken internal links on some pages

serg:


themes/kubermatic-docs/layouts/_default/_markup/render-image.html
```diff
@@ -1,10 +1,10 @@
...
 <figure>
-  <img src="{{ .Destination | safeURL }}" title="{{ .Title }}">
+  <img src="{{ .Destination | safeURL }}" title="{{ .Title }}"{{with .Text}} alt="{{.}}"{{end}}/>
   <figcaption>
     <h4>{{ .Title }}</h4>
   </figcaption>
 </figure>
 {{ else }}
-<img src="{{ .Destination | safeURL }}" alt="{{ .Text }}">
+<img src="{{ .Destination | safeURL }}"{{with .Text}} alt="{{.}}"{{end}}/>
```
embik, line 9: Minor nit, but I don't think the trailing slash does anything in the `img` tag anymore? Should we really add it?

serg: Yeah, the slash is optional. I think this is my habit to mark self closing tags with slash at end 😄 . Now, removed from here.

embik (approved):
/approve

PR #1201, opened 2022-09-01
Update documentation for using custom OSPs with KubeOne

ahmedwaleedmalik:
Fixes https://github.com/kubermatic/kubeone/issues/2341

Signed-off-by: Waleed Malik <ahmedwaleedmalik@gmail.com>

### DIFF
--- a/content/kubeone/master/architecture/operating-system-manager/usage/_index.en.md
+++ b/content/kubeone/master/architecture/operating-system-manager/usage/_index.en.md
@@ -147,6 +147,12 @@ spec:
             data: echo "hello world"
 ```
 
+{{% notice warning %}}
+Make sure to escape all occurrences of Go text template in your custom
+OperatingSystemProfiles. For example, if you want have `{{.Token}}` in your OSP
+then it should be replaced with ```{{ `{{.Token}}` }}```.
+{{% /notice %}}
+
 **NOTE:** This is just for demonstration purposes and the custom OSPs would most likely be more complicated than this.
 
 2. Create your infrastructure
--- a/content/kubeone/v1.5/architecture/operating-system-manager/usage/_index.en.md
+++ b/content/kubeone/v1.5/architecture/operating-system-manager/usage/_index.en.md
@@ -147,6 +147,12 @@ spec:
             data: echo "hello world"
 ```
 
+{{% notice warning %}}
+Make sure to escape all occurrences of Go text template in your custom
+OperatingSystemProfiles. For example, if you want have `{{.Token}}` in your OSP
+then it should be replaced with ```{{ `{{.Token}}` }}```.
+{{% /notice %}}
+
 **NOTE:** This is just for demonstration purposes and the custom OSPs would most likely be more complicated than this.
 
 2. Create your infrastructure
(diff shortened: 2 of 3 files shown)


content/kubeone/master/architecture/operating-system-manager/usage/_index.en.md
```diff
@@ -147,6 +147,9 @@ spec:
             data: echo "hello world"
 ```
 
+{{% notice warning %}} Make sure to escape all occurrences of Go text template in your custom OperatingSystemProfiles. For example, if you want have "{{ .Token }}" in your OSP then it should be replaced with "{{ \`{{ .Token }}\` }}".
```
kron4eg, line 150:
```suggestion
{{% notice warning %}}

Make sure to escape all occurrences of Go text template in your custom
OperatingSystemProfiles. For example, if you want have `{{.Token}}` in your OSP
then it should be replaced with ```{{ `{{.Token}}` }}```.

{{% /notice %}}
```

Not marking it as a code looks a bit meh.
image

New look:
image

content/kubeone/v1.5/architecture/operating-system-manager/usage/_index.en.md
```diff
@@ -147,6 +147,9 @@ spec:
             data: echo "hello world"
 ```
 
+{{% notice warning %}} Make sure to escape all occurrences of Go text template in your custom OperatingSystemProfiles. For example, if you want have "{{ .Token }}" in your OSP then it should be replaced with "{{ \`{{ .Token }}\` }}".
```
kron4eg, line 150: here too

kron4eg (commented):
a bit of formatting improvement could make it even better

content/kubeone/master/architecture/operating-system-manager/usage/_index.en.md
ahmedwaleedmalik: Awesome, I tried doing this and failed, so just gave up(too early :P) :D

kron4eg:
I'm thinking, what about if we provide an awk command / script that would make those replacements for the end user?

kron4eg:
> I'm thinking, what about if we provide an awk/sed command / script that would make those replacements for the end user?

something like
```
sed 's|{{\(.*\)}}|{{ `{{\1}}` }}|g' -i osp-amzn2.yaml
```

but .* is too greedy and need to be fixed somehow else

ahmedwaleedmalik:
> > I'm thinking, what about if we provide an awk/sed command / script that would make those replacements for the end user?
> 
> something like
> 
> ```
> sed 's|{{\(.*\)}}|{{ `{{\1}}` }}|g' -i osp-amzn2.yaml
> ```
> 
> but `.*` is too greedy and need to be fixed somehow else

We can do that but for now TBH even just documenting this is fine. I don't think this is a feasible solution in the long run. I thought about this a bit and adding a dedicated directory `custom-operating-system-profiles` or something that is not "rendered" in-code is the best solution.

ahmedwaleedmalik:
> sed 's|{{\(.*\)}}|{{ `{{\1}}` }}|g' -i osp-amzn2.yaml

This is fine no? I think it's the safest approach since it takes care of the `{{ range ...}}` etc as well. WDYT?

kron4eg:
> > sed 's|{{(.*)}}|{{ `{{\1}}` }}|g' -i osp-amzn2.yaml
> 
> This is fine no? I think it's the safest approach since it takes care of the `{{ range ...}}` etc as well. WDYT?

no, it's broken, it's "greedy regex". If there is line with
```
{{ .Some }} text {{ .Else }}
```
it will replace it with
```
{{ `{{.Some }} text {{ .Else}}` }}
```

PR #1214, opened 2022-09-23
fix runbook script for yq4

xrstf:
We recently switched to only providing `yq` in version 4 in our Docker images, so this script needed to be adjusted.


xmudrii (approved):
/approve

PR #1215, opened 2022-09-26
Update KKP docs based on the recent feedback

xmudrii:
This PR updates 2.18 to 2.19 upgrade document based on the recent feedback. This PR adds:

- note about removing the old Loki StatefulSet when upgrading the `logging/loki` Helm chart
- note about removing the user cluster MLA Prometheus deployment

/assing @wurbanski 

/hold
The following questions need to be answered first:

- Do we also need to remove the Promtail DaemonSet?
- Where we could add a note about the mla-secrets Helm chart?
- Do we need those notes in other upgrade documents?

Additionally, we need to copy this document to other versions.


wurbanski:
> * Do we also need to remove the Promtail DaemonSet?

I think it's the same situation, yes.

> * Where we could add a note about the mla-secrets Helm chart?

`tutorials-howtos/monitoring-logging-alerting/user-cluster/admin-guide/#installation` should contain this information, perhaps it's not clear enough to not deploy it when upgrading :(

The deployment script does not deploy the secrets for this reason as well.

> * Do we need those notes in other upgrade documents?

this change might be required when upgrading to usercluser mla versions 0.1.1+ and 0.2.1+ from older ones. These relate to 2.19.x and 2.20.x KKP releases, accordingly. As such, I think the note about clearing deployments/daemonsets could be useful, we just didn't include notes regarding MLA in the general upgrade notes before, which was a mistake.

xmudrii:
@wurbanski I did some changes based on your comment. PTAL at the PR again.

wurbanski (commented):
/lgtm
/approve

PR #1233, opened 2022-10-25
API server access control page updates

sachintiptur:
Signed-off-by: Sachin Tiptur <sachin@kubermatic.com>
This PR renames and updates the API server network policies page with recent changes, to configure source IP ranges 
to restrict the user cluster api server access from the outside.

### DIFF
--- a/content/kubermatic/main/tutorials-howtos/networking/apiserver-policies/_index.en.md
+++ b/content/kubermatic/main/tutorials-howtos/networking/apiserver-policies/_index.en.md
@@ -43,9 +43,11 @@ spec:
 Please note that this procedure does not affect already running user clusters, for those the API Server Network Policies need to be disabled individually as described in the previous section.
 
 ## API Server Allowed Source IP ranges
-Since KKP v2.22, it is possible to restrict the access to the user cluster API server based on the source IP ranges. To restrict the server access from the internet or to have a limited access,`apiServerAllowedIPRanges` need to be configured. It is also important to note that you allow IP range of the worker nodes too, otherwise worker nodes will not be able to connect to the api server.
+Since KKP v2.22, it is possible to restrict the access to the user cluster API server based on the source IP ranges. By default, the access is not restricted and the API server is accessible from anywhere. To restrict the API server access, Cluster's `spec.apiServerAllowedIPRanges` needs to be configured with the list of allowed IP ranges. Access from any other IP ranges will be denied.
 
-In the cluster spec, set the `apiServerAllowedIPRanges` as shown in below example, 
+When restricting access to the API server, it is important to allow the IP range of the worker nodes network too, otherwise the worker nodes will not be able to connect to the it.
+
+To restrict the access to the API server, set the `apiServerAllowedIPRanges` in the in the cluster spec, as shown in the example below:
 
 ```yaml
 spec:


content/kubermatic/main/tutorials-howtos/networking/apiserver-policies/_index.en.md
```diff
@@ -39,3 +40,16 @@ spec:
 ```
 
 Please note that this procedure does not affect already running user clusters, for those the API Server Network Policies need to be disabled individually as described in the previous section.
+
+## API Server Allowed Source IP ranges
+Since KKP v2.22, it is possible to restrict the access to the user cluster API server based on the source IP ranges. To restrict the server access from the internet or to have a limited access, `front-loadbalancer` k8s service can be used to allow only the configured IP source ranges. It can be confiured as `.spec.loadBalancerSourceRanges`. This works only when expose strategy is set to `LoadBalancer`.
```
rastislavs, line 45: I would maybe avoid these sentences, as it describes KKP internals and is not so important for the users who just want to use the feature - and may even mislead them:

> To restrict the server access from the internet or to have a limited access, `front-loadbalancer` k8s service can be used to allow only the configured IP source ranges. It can be confiured as `.spec.loadBalancerSourceRanges`

content/kubermatic/main/tutorials-howtos/networking/apiserver-policies/_index.en.md
```diff
@@ -39,3 +40,16 @@ spec:
 ```
 
 Please note that this procedure does not affect already running user clusters, for those the API Server Network Policies need to be disabled individually as described in the previous section.
+
+## API Server Allowed Source IP ranges
+Since KKP v2.22, it is possible to restrict the access to the user cluster API server based on the source IP ranges. To restrict the server access from the internet or to have a limited access, `front-loadbalancer` k8s service can be used to allow only the configured IP source ranges. It can be confiured as `.spec.loadBalancerSourceRanges`. This works only when expose strategy is set to `LoadBalancer`.  
+
+In the cluster spec, set the `apiServerAllowedIPRanges` as shown in below example,
```
rastislavs, line 47: Please also mention, that when restricting access to apiserver, it is very important to also allow IP range of the worker nodes, otherwise they will not be able to connect to the apiserver.

rastislavs (changes requested):

content/kubermatic/main/tutorials-howtos/networking/apiserver-policies/_index.en.md
```diff
@@ -39,3 +41,16 @@ spec:
 ```
 
 Please note that this procedure does not affect already running user clusters, for those the API Server Network Policies need to be disabled individually as described in the previous section.
+
+## API Server Allowed Source IP ranges
+Since KKP v2.22, it is possible to restrict the access to the user cluster API server based on the source IP ranges. To restrict the server access from the internet or to have a limited access,`apiServerAllowedIPRanges` need to be configured. It is also important to note that you allow IP range of the worker nodes too, otherwise worker nodes will not be able to connect to the api server.
```
rastislavs, line 46:
```suggestion
Since KKP v2.22, it is possible to restrict the access to the user cluster API server based on the source IP ranges. By default, the access is not restricted and the API server is accessible from anywhere. To restrict the API server access, Cluster's `spec.apiServerAllowedIPRanges` needs to be configured with the list of allowed IP ranges. Access from any other IP ranges will be denied.

When restricting access to the API server, it is important to allow the IP range of the worker nodes network too, otherwise the worker nodes will not be able to connect to the it.
```

content/kubermatic/main/tutorials-howtos/networking/apiserver-policies/_index.en.md
```diff
@@ -39,3 +41,16 @@ spec:
...
+
+In the cluster spec, set the `apiServerAllowedIPRanges` as shown in below example,
```
rastislavs, line 48:
```suggestion
To restrict the access to the API server, set the `apiServerAllowedIPRanges` in the in the cluster spec, as shown in the example below:
```

PR #1241, opened 2022-11-11
Update swagger specs from patch releases 2.21.3, 2.20.10 and 2.19.12

embik:
Swagger specs are outdated in our docs because they're only synced automatically up to the initial minor release. This bumps the files to what is on `cmd/kubermatic-api/swagger.json` for the respective release tags.


ahmedwaleedmalik (approved):
/approve

PR #1243, opened 2022-11-14
Copy over vSphere needed permissions for KKP

moadqassem:
It seems that, we kinda forget to copy over the correct and full required permissions for vSphere cloud provider for KKP 2.21.0 and main branch. This PR copies those over.


ahmedwaleedmalik (approved):
/approve

PR #1251, opened 2022-11-15
Add migration guide for kubevirt VMs regarding cluster labels

dermorz:
This contains the migration steps for already existing KubeVirt VMs after upgrading to KKP 2.22.

Reference: https://github.com/kubermatic/kubermatic/issues/10923

### DIFF
--- a/content/kubermatic/main/architecture/supported-providers/kubevirt/kubevirt.en.md
+++ b/content/kubermatic/main/architecture/supported-providers/kubevirt/kubevirt.en.md
@@ -554,6 +554,59 @@ Follow the below steps to import the dashboard in Grafana:
 
 ![Grafana Dashboard](/img/kubermatic/main/monitoring/kubevirt/grafana.png)
 
-## Breaking Changes
+## Migration to KKP 2.22
 
-Please be aware that between KKP 2.20 and KKP 2.21, a breaking change to the `MachineDeployment` API for KubeVirt has occurred. For more details, please [check out the 2.20 to 2.21 upgrade notes]({{< ref "../../../tutorials-howtos/upgrading/upgrade-from-2.20-to-2.21/#kubevirt-migration" >}}).
+### Cluster labels
+
+With KKP 2.22 KubeVirt CCM is upgraded from `v0.2.0` to `v0.4.0`.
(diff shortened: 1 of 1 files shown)


content/kubermatic/main/architecture/supported-providers/kubevirt/kubevirt.en.md
```diff
@@ -554,6 +555,59 @@ Follow the below steps to import the dashboard in Grafana:
 
 ![Grafana Dashboard](/img/kubermatic/main/monitoring/kubevirt/grafana.png)
 
-## Breaking Changes
+## Migration to KKP 2.22
```
mfranczy, line 558: please add as well a new section:
````
### KubeVirt upgrade 
````
it's just a place that we don't forget about it :) later I will fill this section after @sankalp-r 's is merged.

File: content/kubermatic/main/architecture/supported-providers/kubevirt/kubevirt.en.md [hunk 2]
```diff
@@ -554,6 +555,73 @@ Follow the below steps to import the dashboard in Grafana:
...
+
+### MachineDeployments
+
+MachineDeployments created with KKP 2.22 will have the field `clusterName` set with the cluster ID in the KubeVirt `cloudProviderSpec`,
+which is used by the machine-controller to set the cluster labels on VMs and VMIs. Any pre-existing MachineDeployments will not have this
+field set automatically to avoid rolling over all Machines. So in the scenario of scaling up pre-existing MachineDeployments for
+LoadBalancers to properly function, there are 2 possibilities:
+
+1. If it's no problem that all VMs get re-created, just set `MachineDeployment.spec.template.spec.providerSpec.value.cloudProviderSpec.clusterName` to your cluster ID.
+2. Otherwise you need to manually set the cluster labels described above on all VMs and VMIs that are being created while scaling the MachineDeployment up.
```
hdurand0710, line 627: Do we want to tell the customer to also update the MD (will be useful in case of md scaling for example) ?
And maybe warn him that this will imply that a new VM will be spawned ?
What do you think ?

mfranczy:
I would like to strutcture this a little bit different. By step-by-step example you assume that users will already use manual methods for machine deployments. So giving them choice to edit machine deployment and add cluster name... the first step is not necessary. We should be a little bit more explicit.

I will be able to propose something tomorrow, today i have a full stack.. :)

PR #1252, opened 2022-11-15
Add notice about upgrading and mla-secrets chart.

wurbanski:
There were a lot of reports of issues with credentials after reinstalling mla-secrets during upgrade procedures. This adds an additional notice to not do that and changes the default installation command so that it won't upgrade the chart by default.


archups (approved):
LGTM

PR #1256, opened 2022-11-18
Added documentation for ApplicationDefinition defaultValues

archups:
Signed-off-by: archups <archupsawant@gmail.com>


xrstf (changes requested):
I think this is backwards. These files in this repository will be overwritten once the next PR is merged in KKP's main branch (here: https://github.com/kubermatic/kubermatic/blob/main/hack/ci/update-docs.sh#L45-L52).

If you need more documentation comments, adjust https://github.com/kubermatic/kubermatic/blob/main/codegen/example-yaml/main.go#L284 to ensure the defaultValues are not empty, so that they are not skipped when generating the YAML (just run `hack/update-docs.sh` and check out the changes to the YAML).

toschneck:
@ezasewe please follow up on this as @archups is OOO

PR #1266, opened 2022-11-28
Add documentation for defaulting tag category in the seed

moadqassem:
Document the tag category defaulting in seed objects.


xmudrii:
@moadqassem @embik Shouldn't we also update the main documentation? I believe this will also be part of 2.22 or is that not the case?

embik:
> @moadqassem @embik Shouldn't we also update the main documentation? I believe this will also be part of 2.22 or is that not the case?

Yes please. I missed that.

PR #1273, opened 2022-12-06
Fix typo

ofek:
Signed-off-by: Ofek Lev <ofekmeister@gmail.com>

PR #1274, opened 2022-12-06
Fix KubeOne tutorial typo on main as well

embik:
#1273 fixed a typo on the publicly visible 1.5 tutorial page for KubeOne, I forgot to check if it needed fixing for `main` as well. Turns out it did. This is the PR to do that.


ahmedwaleedmalik (approved):
/approve
yesyoucan

PR #1277, opened 2022-12-07
Update support matrix for Kubernetes 1.23 support in KKP 2.20

embik:
We forgot to update this ("we" as in "I"). This PR updates the docs to include Kubernetes 1.23 support in KKP 2.20, which we added in kKP 2.20.3.


ahmedwaleedmalik (approved):
/approve

PR #1281, opened 2022-12-09
Document an issue with upgrading Canal CNI in KubeOne

xmudrii:
Document an issue with upgrading Canal CNI in KubeOne. See https://github.com/kubermatic/kubeone/pull/2538 for more details.


ahmedwaleedmalik (approved):
/approve

PR #1294, opened 2023-01-10
Add TOC sidebar

### LINKED ISSUE #1150: Add floating TOC on right side of content
After some [user feedback](https://github.com/kubermatic/dashboard/issues/4737#issuecomment-1210599173) on our docs, I discussed having a floating TOC with @cschieder and we agreed this might be useful. This can help users discover content on a page (we don't always set `enableTOC` on our pages and that is only at the top of a page anyway) and improve orientation/navigation while going over a longer document. Usually, this "floating TOC" is on the right side of the actual content and frames it together with the navigation on the left.

Examples for this kind of "floating TOC" (not sure if this is the proper web development term for it) are the AWS docs, for example [this page](https://docs.aws.amazon.com/eks/latest/userguide/kubernetes-versions.html). Kubernetes [also does it](https://kubernetes.io/docs/concepts/overview/what-is-kubernetes/).

Ideally, this would:

- Highlight the section that a user is currently looking at
- Be collapsible, so users can focus on the content if they don't need the additional navigation help
- Maybe not show for pages that don't have additional headlines (minor/small pages)

But those are all "nice to have" in my view and not required.

serg:
fix #1150


content/kubermatic/main/installation/start-kkp/guides/wizard/_index.en.md
```diff
@@ -0,0 +1,109 @@
++++
```
embik, line 1: I think this file shouldn't be here? I suspect it existed when this PR was started but removed in the mean time. Can you remove it again please?

content/kubermatic/main/installation/start-kkp/guides/setup-repository/_index.en.md
```diff
@@ -0,0 +1,224 @@
++++
```
embik, line 1: Same here.

embik (changes requested):
Minor change needed, but looking pretty good!

embik (approved):
/approve

PR #1297, opened 2023-01-12
Add `oidc_logout_url` in OIDC provider configuration docs

Waseem826:
A new `oidc_logout_url` endpoint was added in https://github.com/kubermatic/dashboard/pull/5521. If this endpoint is provided then UI will redirect to it after logging out.


ahmedwaleedmalik (approved):
/approve

PR #1312, opened 2023-01-23
Update KKP component versions for all supported versions

embik:
These pages have been outdated for quite a while, this PR brings them up to what we actually ship. The PR removes the "nice" markdown formatting because eventually I want this to be automated. And it still renders correctly in our docs anyway.

For future reference, I used the following shell thing to generate this:

```sh
for filepath in $(find charts -name Chart.yaml | sort); do echo "| $(echo ${filepath} | gsed -e 's/^charts\///g' -e 's/\/Chart\.yaml$//g') | $(yq '.appVersion' ${filepath} | gsed -e 's/^v//g') |"; done
```

And then manually updated the versions for `kubermatic-operator`.


embik:
For future reference, this has been automated for `main` via https://github.com/kubermatic/kubermatic/pull/11731.

PR #1314, opened 2023-01-24
Update headers to be more consistent in language

embik:
Many headers in our docs were not following consistent language and title casing. This PR tries to make them more consistent, remove headers that are duplicating information, improve the header structure, etc. A few minor content improvements snuck in while I was going over all pages too.


vgramer (commented):
love this improvement!
/lgtm

PR #1331, opened 2023-02-01
Fix brew install command in KKP install tutorials

xmudrii:
If I'm not mistaken, `brew install gnu-tar` is supposed to install the GNU version of `tar`, but we assume in docs it installs the GNU version of `tr` (`gtr`). `gtr` is supposed to be installed with the `coreutils` formula, so this PR adjusts our docs to recommend installing that formula instead.

/assign @embik


embik (approved):
good catch!

PR #1348, opened 2023-02-07
Add some clarifications for expose strategy migration

embik:
This PR adds two notes to the expose strategy migration:

1. A recommendation to terminate all workloads prior to migration. As Kubernetes won't be able to communicate with kubelets on any nodes, they cannot be properly shut down after the migration. That seems ... dangerous at times. Users should exercise control about that shutdown process in my opinion.
2. A note saying that dead `Nodes` need to be removed.

This is a result of testing (https://github.com/kubermatic/kubermatic/issues/11831).


rastislavs:
/lgtm
/approve

Thanks ;)

Issue #1302, opened 2023-01-19
Deployment: time, keys, data storage and accounts


csengerszabo:
Installation and configuration part of the docs need to be updated with new paragraphs:

- [x] Installation: Specify the expected amount of time it takes to complete deployment in the introductory material / deployment guide. Let's define with the master cluster installation and define another time with assuming that the master cluster installation is automated (can be installed from a cloud store), but seeds, OIDC, etc. should be still configured.
- [x] Installation and Configuration: Describe the purpose and location of each key the user is instructed to create during the installation
- [x] Installation: some extended information about the needed accounts and account restrictions (permission control) for installing and running KKP in the environment.
- [x] Installation: about details on where customer sensitive data are stored (what parts of the architecture)
- [x]  Move this https://github.com/kubermatic/kubeone/tree/main/examples/terraform/aws configuration documentation to docs
- [x]  Add this easier architecture picture: https://docs.google.com/presentation/d/1HpAEaJBMDrxA88p710DsauxGmZDKf2rL2c97Qsgs1ig/edit#slide=id.g1990739d9ad_1_44
- [x]  Guidance on managing AWS service limits in the AWS parts of the documentation. E.g.: redirecting to AWS Documentation on service limits (https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-resource-limits.html) where relevant service involved in our solution.

INT-004, DSEC-007, RM-004

PR #1349, opened 2023-02-07
refine installation guide

SimonTheLeg:
see #1302


SimonTheLeg:
@embik This PR covers a couple of topics simultaneously (for all the details see the attached issue), but there is one thing I wanted your opinion on: The best way I found to make the connection between the KKP installation guide and our existing reference on required [infrastructure permissions](https://docs.kubermatic.com/kubeone/v1.5/architecture/requirements/machine-controller/), was to include the presets guide as a Next steps in the installation guide and then ammend the presets guide with a link to the infra permissions reference. This was the best flow I could figure out. Since you have done the rewrite of the install guide: How do you feel about this?

content/kubeone/main/architecture/requirements/machine-controller/_index.en.md
```diff
@@ -6,3 +6,5 @@ weight = 20
 +++
 
 See detail pages for specific requirements of the cloud providers used by the Kubermatic machine-controller.
+
+{{% children %}}
```
embik, line 10: OMG I had no idea this shortcode exists.

content/kubermatic/main/installation/install-kkp-CE/_index.en.md
```diff
@@ -97,7 +107,7 @@ The key items to consider while preparing your configuration files are described
 | To authenticate via an external identity provider you need to set up connectors in Dex. Check out [the Dex documentation](https://dexidp.io/docs/connectors/) for a list of available providers. This is not required but highly recommended for multi-user installations. | `.dex.connectors` (`values.yaml`; not included in example file) |
 | The expose strategy which controls how control plane components of a User Cluster are exposed to worker nodes and users. See [the expose strategy documentation]({{< ref "../../tutorials-howtos/networking/expose-strategies/" >}}) for available options. Defaults to `NodePort` strategy if not set. | `.spec.exposeStrategy` (`kubermatic.yaml`; not included in example file) |
 
-There are many more options but these are essential to get a minimal system up and running. The secret keys
+There are many more options but these are essential to get a minimal system up and running. A full reference of all options can be found in the [KubermaticConfiguration Reference]({{< relref "../../references/crds/_index.en.md#kubermaticconfigurationspec" >}}). The secret keys
```
embik, line 110: You can drop `_index.en.md` here and just make the ref `../../references/crds/#kubermaticconfigurationspec`.

content/kubermatic/main/architecture/_index.en.md
```diff
@@ -7,11 +7,11 @@ weight = 5
 
 Kubermatic Kubernetes Platform (KKP) makes full use of Kubernetes cluster to organize and scale workloads, depending on your and your customer's needs. In a typical small-scale setup, pictured below, a single cluster contains KKP and the master components for every user cluster.
 
-![KKP Architecture Diagram](/img/kubermatic/main/architecture/combined-master-seed.png)
+![KKP Architecture Diagram](/img/kubermatic/main/architecture/high-level-architecture.png)
```
embik, line 10: Does it make sense to drop this diagram completely? It's removed as a file later on, but wouldn't it still be useful to showcase the "combined master/seed" setup?

content/kubermatic/main/installation/install-kkp-CE/_index.en.md
```diff
@@ -21,6 +22,15 @@ migrating existing installations.
 For this guide you need to have [kubectl](https://kubernetes.io/docs/tasks/tools/#kubectl) and [Helm](https://www.helm.sh/) (version 3) installed locally.
 You should be familiar with core Kubernetes concepts and the YAML file format before proceeding.
 
+<!-- TODO Add quotas for additional providers-->
+In addition, we recommend to familiarize yourself with the resource-quota system of your infrastructure provider. It is important to leave free capacity to allow KKP to provision infrastructure for your future Kubernetes clusters, but also to enforce a maximum to protect against overspending.
```
embik, line 26:
```suggestion
In addition, we recommend familiarizing yourself with the resource quota system of your infrastructure provider. It is important to provide enough capacity to let KKP provision infrastructure for your future user clusters, but also to enforce a maximum to protect against overspending.
```

This might be personal taste, but something about this sentence felt odd, so maybe this is better?

embik (changes requested):

content/kubermatic/main/architecture/_index.en.md
SimonTheLeg: Maybe @csengerszabo can chime in here. I remember that it was in the original task to replace the architecture picture, but tbh I am not fully sure what the intention of the new one was

csengerszabo: Can we then somehow display both? The original is the general architecture, and this one is more likely an example

content/kubermatic/main/tutorials-howtos/administration/presets/_index.en.md
```diff
@@ -236,3 +236,8 @@ kubectl edit preset my-preset
 
 This will open the editor configured for `kubectl edit` and let you update the YAML specification.
 It is possible to update any fields already present or add/remove provider sections from the `Preset`.
+
+## Limiting Permissions for Credentials used in Presets
+
+For selected infrastructure providers it makes sense to limit the permissions of the credentials to protect against access creep.
+For these cases, we have compiled a [list of minimum permissions]({{< relref "../../../../../kubeone/main/architecture/requirements/machine-controller/" >}}) required for different providers.
```
embik, line 243: @SimonTheLeg let's discuss your point here:

> @embik This PR covers a couple of topics simultaneously (for all the details see the attached issue), but there is one thing I wanted your opinion on: The best way I found to make the connection between the KKP installation guide and our existing reference on required [infrastructure permissions](https://docs.kubermatic.com/kubeone/v1.5/architecture/requirements/machine-controller/), was to include the presets guide as a Next steps in the installation guide and then ammend the presets guide with a link to the infra permissions reference. This was the best flow I could figure out. Since you have done the rewrite of the install guide: How do you feel about this?

I like the flow you described, however I think linking to KubeOne is not a good idea here. The permissions documented in KubeOne's machine-controller section refer to what is needed to create VMs on respective cloud providers. But KKP needs permissions beyond that for setting up cloud resources per user cluster. So these permissions are likely incomplete. Have you tried them out for working with KKP?

embik (changes requested):

content/kubermatic/main/tutorials-howtos/administration/presets/_index.en.md
SimonTheLeg: So I have not tried them out personally, but asked in the PS channel what we typically use for customers. Maybe @dharapvj can chime in on this: When we talked about it on Slack, is the documentation linked only valid for Kubeone or do we also use this for KKP?

dharapvj: that's a good point.. I would imagine that these same permissions work in both kubeone and kkp since both use machine-controller to create machines. But I have lesser exposure to KKP deployments in AWS in my daily work

embik: That's my point, KKP does use machine-controller to create machines, but it also creates the resources needed to provision those machines. All the stuff happening in https://github.com/kubermatic/kubermatic/tree/main/pkg/provider/cloud, effectively. KubeOne would need some of these permissions for Terraform as well (but are they the same? We don't know), but the linked page is explicitly only for machine-controller, so it should be considered a subset of needed permissions.

dharapvj: Agreed.

Then only way out .. probably is to start with absolute minimal permission IAM user and go on adding permissions until everything has panned out.

Also - see this file - which as some interesting things in it..
https://github.com/kubermatic/kubermatic/blob/main/pkg/provider/cloud/aws/policy.go

csengerszabo (commented):
- I couldn't find the part where details on where customer sensitive data are stored, can you point that out?
- Is there a reference to https://github.com/kubermatic/kubeone/tree/main/examples/terraform/aws documentation ?

SimonTheLeg:
@csengerszabo 

> I couldn't find the part where details on where customer sensitive data are stored, can you point that out?

For now I have kept it generic and said that infrastructure provider credentials are stored on the master and seed clusters. Apart from this there’s not much meaningful I could add; they are not really stored in any special place. They are just inside the etcd database of these clusters.

image


> Is there a reference to [kubermatic/kubeone@main/examples/terraform/aws](https://github.com/kubermatic/kubeone/tree/main/examples/terraform/aws?rgh-link-date=2023-02-20T14%3A23%3A28Z) documentation ?

This was basically already in place: The KKP install guide leads you to use kubeone (which is also the only install method where the example is valid) and then that install guide directly links to the examples per provider

content/kubermatic/main/architecture/_index.en.md
SimonTheLeg: I have pushed a proposal how we can do this now:

@embik For now I have worded it with “the master cluster” and then used plural for seed- and user-clusters. Is this correct, or is there a possibility for a customer to use more than one master and it is still considered a single KKP installation?

@csengerszabo I have noticed that in the high-level diagram the "master cluster" is called "root cluster". Is this on purpose? If not we should change it in the diagram to have consistent naming. I can do this, but I would need write permissions on the Google Doc to change the diagram.

embik:
> @csengerszabo
> 
> > I couldn't find the part where details on where customer sensitive data are stored, can you point that out?
> 
> For now I have kept it generic and said that infrastructure provider credentials are stored on the master and seed clusters. Apart from this there’s not much meaningful I could add; they are not really stored in any special place. They are just inside the etcd database of these clusters.

I would actually like to disagree on that; the information which Kubernetes resources will hold sensitive data can be used to configure encryption-at-rest, a regulatory or certification requirement that KKP admins might have. I think that KKP can store sensitive data not only in `Secrets`, but also `Presets` (and maybe other resources?), so I think the resources (and their location) for this should be enumerated.

content/kubermatic/main/architecture/_index.en.md
embik: Another question on this diagram, it mentions "Backup to A/B/C" for user clusters. This seems to suggest to me that you can back up / fall over to other seeds (because they are also named A/B/C), which is not the case. So I'm not sure what the diagram is saying here?

csengerszabo (approved):
/lgtm

PR #1356, opened 2023-02-13
Document Ubuntu 22.04 support since KKP 2.21.2

embik:
Support for Ubuntu 22.04 was added in https://github.com/kubermatic/kubermatic/pull/11072. This PR updates docs accordingly.


ahmedwaleedmalik (approved):
/approve

PR #1379, opened 2023-02-21
Add upgrade guide and document known issues for KubeOne 1.6

xmudrii:
Add upgrade guide and document known issues for KubeOne 1.6.

/assign @kron4eg @ahmedwaleedmalik

### DIFF
--- a/content/kubeone/main/known-issues/_index.en.md
+++ b/content/kubeone/main/known-issues/_index.en.md
@@ -19,7 +19,7 @@ please consult the [appropriate changelog][changelogs].
 
 |              |                                                   |
 |--------------|---------------------------------------------------|
-| Status       | Fixed in KubeOne 1.6.0-rc.2                       |
+| Status       | Fixed in KubeOne 1.6.0                            |
 | Severity     | Critical                                          |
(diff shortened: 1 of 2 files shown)


content/kubeone/main/known-issues/_index.en.md
```diff
@@ -1,138 +1,118 @@
...
 
-Clusters running Calico VXLAN might not be able to reach ClusterIP Services
-from a node where the pod is running.
-
-### Recommendation
+Users who:
 
-**We do NOT recommend upgrading to KubeOne 1.5 at this time if you're using
-Calico VXLAN. Follow the linked GitHub issue and this page for updates.**
+- used **KubeOne 1.6.0-rc.1** or built KubeOne manually on
```
kron4eg, line 28: Probably it's better to put accent on the fact that the fix is available after _1.6.0-rc.1 or later_, with emphasis on later (having a mind a release version), otherwise it looks like as if we are recommending an RC pre-release version over the release (yes, at this point it's future release, but still).

kron4eg: OK, scratch that, I've read it wrongly

PR #1388, opened 2023-02-23
Update KubermaticConfiguration for v2.22

embik:
We had to update `KubermaticConfiguration` on short notice, so this syncs the data for v2.22 accordingly.


xrstf (approved):
/approve

PR #1394, opened 2023-02-24
Add KKP release notes in our docs

embik:
This adds our changelog as release notes to docs.kubermatic.com. This was an idea by @ahmedwaleedmalik.

cc @xmudrii @csengerszabo for opinions on this.

### DIFF
--- a/content/kubermatic/v2.20/how_to_contribute_to_kkp/_index.en.md
+++ b/content/kubermatic/v2.20/how_to_contribute_to_kkp/_index.en.md
@@ -1,7 +1,7 @@
 +++
 title = "Contribute to KKP"
 date = 2021-04-24T15:00:15+02:00
-weight = 61
+weight = 80
 +++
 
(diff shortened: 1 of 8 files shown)


csengerszabo:
Good idea. Only question I have in mind: should we put this in the root or rather in the architecture/ section? Since all compatibility and versioning related topics are elaborated in the latter.

embik:
> Good idea to add it. Only question I have in mind: should we put this in the root or rather in the architecture/ section? Since all compatibility and versioning related topics are elaborated in the latter.

We discussed this on the sig-documentation channel and people specifically requested it to be in the root directory. I don't have strong opinions on this in any case.

xmudrii:
I agree that adding it to the root directory is perhaps the best place for changelogs. Also, if we do this for KKP, we should also do it for KubeOne.

csengerszabo (approved):
/lgtm

PR #1403, opened 2023-03-07
Add child pages with depth to top-level pages

embik:
This adds child pages with depth 5 to all top-level pages (Architecture, Installation, etc).

In my opinion, this can help discover pages in a tree view, but I'd like to hear some thoughts on it.


csengerszabo (approved):
This is dynamic and structured, I really like it.

/lgtm

PR #1408, opened 2023-03-16
Fix multiline issue for notice shortcode

serg:
Example: https://docs.kubermatic.com/kubermatic/v2.22/installation/upgrading/upgrade-from-2.19-to-2.20/
/assign @embik


embik (approved):
/approve

PR #1416, opened 2023-03-28
clarify that we need no AWS root account access

SimonTheLeg:
Makes the no root more apparent to satisfy DSEC-002


csengerszabo (approved):
/lgtm

PR #1428, opened 2023-03-30
clarify that sensitive information is stored in etcd

SimonTheLeg:
This should address DSEC-008. It clarifies that the data is stored in etcd and also backports that change to the v2.22 docs as requested in the meeting today.


SimonTheLeg:
/hold

so both reviewers can have a look and it does not get merged accidentally

PR #1429, opened 2023-03-30
explicitedly state support for multi-zone

SimonTheLeg:
(optional) We can explitiedly say that some of our examples include options for multi-zone deployments for INT-003, if a link to our examples is not enough


csengerszabo (approved):
/lgtm

PR #1431, opened 2023-04-05
Update _index.en.md

hc2p:
Added small note about taints for k8s 1.25+


embik:
This file is auto-synced from KubeOne code (see https://github.com/kubermatic/docs/tree/master/content/kubeone/main/references/kubeone-cluster-v1beta2 and the commits on it), so I think we should add this note in the source material and not here.

PR #1435, opened 2023-04-26
Refresh CRD documentation from latest KKP patch releases

embik:
We haven't updated the CRD reference in a while. This bumps the CRD documentation to the content generated from KKP 2.22.2, 2.21.8 and 2.20.15.


xrstf:
Why are we doing this manually?

embik:
Because I just opened https://github.com/kubermatic/kubermatic/issues/12200 for someone to pick this up for future releases, but I also wanted to fix the situation right now.

xrstf (approved):
/approve

PR #1436, opened 2023-04-27
Update sample manifests for KubeOne cluster creation

ahmedwaleedmalik:
We should, right from the start, instruct our users to use external CCM whenever it's available.

### DIFF
--- a/content/kubeone/main/tutorials/creating-clusters/_index.en.md
+++ b/content/kubeone/main/tutorials/creating-clusters/_index.en.md
@@ -900,16 +900,17 @@ cloudProvider:
 
     [Network]
     public-network = "NAT Network"
-  csiconfig: |
+  csiConfig: |
     [Global]
+    # Unique identifier; name of the cluster
     cluster-id = "<CLUSTER-ID>"
-    cluster-distribution = "<CLUSTER-DISTRIBUTION>"
-
-    [VirtualCenter "<VCENTER-ADDRESS>"]
-    insecure-flag = "false"
     user = "<USERNAME>"
     password = "<PASSWORD>"
     port = "<PORT>"
+    insecure-flag = "false"
+
+    [VirtualCenter "<VCENTER-ADDRESS>"]
+    # List of comma separated datacenters where node VMs are present
     datacenters = "<DATACENTER>"
 ```
 
--- a/content/kubeone/main/tutorials/creating-clusters/_index.en.md
+++ b/content/kubeone/main/tutorials/creating-clusters/_index.en.md
@@ -608,6 +608,8 @@ supported provider.
 
 {{< tabs name="Manifests" >}}
 {{% tab name="AWS" %}}
+`external: true` instructs KubeOne to deploy the
+[AWS Cloud Controller Manager](https://github.com/kubernetes/cloud-provider-aws).
 
 ```yaml
 apiVersion: kubeone.k8c.io/v1beta2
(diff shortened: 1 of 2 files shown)


kron4eg:
/lgtm

@xmudrii PTAL

content/kubeone/main/tutorials/creating-clusters/_index.en.md
```diff
@@ -861,9 +875,10 @@ automatically by KubeOne as of v1.0.4.**
 apiVersion: kubeone.k8c.io/v1beta2
 kind: KubeOneCluster
 versions:
-  kubernetes: '1.22.5'
+  kubernetes: '1.25.6'
 cloudProvider:
   vsphere: {}
+  external: true
```
xmudrii, line 881: For vSphere we also need `csiConfig` or otherwise validation is going to fail.

xmudrii (changes requested):

content/kubeone/main/tutorials/creating-clusters/_index.en.md
ahmedwaleedmalik: Good catch, that's actually very important.

content/kubeone/main/tutorials/creating-clusters/_index.en.md
```diff
@@ -885,6 +900,17 @@ cloudProvider:
 
     [Network]
     public-network = "NAT Network"
+  csiconfig: |
```
xmudrii, line 903: It should be `csiConfig`:

```suggestion
  csiConfig: |
```

content/kubeone/main/tutorials/creating-clusters/_index.en.md
```diff
@@ -885,6 +900,17 @@ cloudProvider:
...
+    [Global]
+    cluster-id = "<CLUSTER-ID>"
+    cluster-distribution = "<CLUSTER-DISTRIBUTION>"
```
xmudrii, line 906: Can we document how to set those two values? The first value should be the name of the cluster, but I'm not sure about the second value.

content/kubeone/main/tutorials/creating-clusters/_index.en.md
```diff
@@ -885,6 +900,17 @@ cloudProvider:
 
     [Network]
     public-network = "NAT Network"
+  csiconfig: |
+    [Global]
```
xmudrii, line 904: Don't we need credentials here as well? We have credentials here in E2E tests for example. Also, do we need parameters such as default-datastore and folder?

xmudrii (changes requested):
Some more comments related to vSphere.

content/kubeone/main/tutorials/creating-clusters/_index.en.md
ahmedwaleedmalik: I basically picked this up from our kubeone init docs :D But let me take a look at this stuff.

content/kubeone/main/tutorials/creating-clusters/_index.en.md
ahmedwaleedmalik: Credentials are already there. Folder and datastore are something that we use but are not a requirement. 

The configuration in the documentation looks fine now. We may want to document other options in the CSI config but that's out of scope for this PR.

xmudrii (approved):
Thank you!
/lgtm
/approve

PR #1459, opened 2023-06-09
Exclude CentOS stream versions from support matrix

wozniakjan:
The claim for CentOS support matrix was worded conclusively. It wasn't clear what CentOS versions are supported and what aren't.

The CentOS versioning is split between Linux and Stream types. KKP intended to support some CentOS Linux 7 (starting with 7.4) and CentOS Linux 8 but didn't intend to promise support for CentOS Stream 8 or CentOS Stream 9.


xrstf (changes requested):
Please give a description as to _why_ this is suddenly excluded.

wozniakjan:
@xrstf good point, described my understanding of the situation, ptal.

xrstf (approved):
/approve

PR #1464, opened 2023-06-14
Update crd-ref-docs to v0.0.9

embik:
https://github.com/elastic/crd-ref-docs/releases/tag/v0.0.9 is out now, so we should upgrade to that. the most important change is that embedded structs get resolved correctly. They are an implementation detail of our API types and users do not care about them, they only want to know what fields are available.


xrstf (approved):
/approve

PR #1465, opened 2023-06-15
Add docs for share cluster admin setting

Waseem826:


Waseem826:
/hold
Need to check if some other docs should also be updated.

ahmedwaleedmalik (approved):
/approve

PR #1469, opened 2023-06-19
Update outdated docs and fix broken images

Waseem826:
- Updated Groups docs
- Updated order of KubeOne provider pages
- Removed `height` from query param of image links to avoid images looking like stretched or too small.


ahmedwaleedmalik (approved):
/approve

PR #1473, opened 2023-06-20
Fixed details on role-definitions

hc2p:
* Fixed incorrect indentation
* Added missing permissions to the list when they were mentioned in the govc output (checked in customer project if valid)


embik (approved):
/approve

PR #1474, opened 2023-06-22
Make Seed config a sub-chapter of 'Custom Configuration'

embik:
This moves the Seed configuration under "Custom Configuration", since it being under "UI Configuration" doesn't make sense.


pkprzekwas:
Good catch, it should be `###` from the very beginning.

/lgtm

Issue #1478, opened 2023-06-22
Document `kubermatic-installer local kind` command


wozniakjan:
https://github.com/kubermatic/kubermatic/pull/12216 introduced new command for the installer
```
kubermatic-installer local kind
```
Which has the main purpose of creating `kind` cluster and installing KKP there + auto-configuring KubeVirt for user clusters.

It is desired to document the feature and explain some shortcomings as discussed in https://github.com/kubermatic/kubermatic/issues/12340, especially:
* `inotify` for KubeVirt: https://github.com/kubermatic/kubermatic/issues/12340#issuecomment-1602433880
* adblocker for `nip.io`: https://github.com/kubermatic/kubermatic/issues/12340#issuecomment-1602368805

PR #1484, opened 2023-06-26
Add installer local command

wozniakjan:
closes: https://github.com/kubermatic/docs/issues/1478

### DIFF
--- a/content/kubermatic/main/installation/local-kkp-installation/_index.en.md
+++ b/content/kubermatic/main/installation/local-kkp-installation/_index.en.md
@@ -1,0 +1,95 @@
++++
+title = "Local Kubermatic Kubernetes Platform (KKP) CE Installation"
+linkTitle = "Local Installation"
+date = 2023-06-26T09:49:10+02:00
+weight = 200
+enableToc = true
++++
+
+{{% notice warning %}}
+Local KKP installation is **not** intended for production setups.
+{{% /notice %}}
+
+This page will guide you through using the KKP installer local command `kubermatic-installer local`. This command simplifies and automates the installation of KKP CE using `kind` and a preconfigured KubeVirt seed. The command is intended only for evaluation and local development purposes. For production KKP installation use the [CE installation guide](../install-kkp-ce) or [EE installation guide](../install-kkp-ee).
+
+## Pre-Installation Requirements
+
+- **Operating System:** Currently, only Linux is supported.
+- **Disk Space:** A minimum of 10Gi is recommended.
+- **RAM:** At least 8Gi of RAM is needed; 16Gi is recommended for more than a single KubeVirt user cluster.
+- **`kind`** (version 0.17 or higher): Please refer to the [`kind` documentation](https://kind.sigs.k8s.io/docs/user/quick-start/) for installation instructions.
+
+{{% notice warning %}}
+The `kubermatic-installer local kind` is currently considered experimental. It will install [KubeVirt](https://kubevirt.io/quickstart_kind/) and [CDI](https://kubevirt.io/labs/kubernetes/lab2.html) inside of the `kind` cluster and for every user cluster download VM image from Kubermatic hosted registry. KubeVirt nodes may require significant portion of your available CPU and RAM, in addition to 600Mi of disk space per VM image.
+{{% /notice %}}
+
+## Installation Procedure
+
+Follow these steps to use the KKP installer local command:
+
+**1. Download the installer.** This is the only manual step; there's no need to prepare any configuration since the installer should automatically configure KKP.
+
+```bash
+VERSION=$(curl -w '%{url_effective}' -I -L -s -S https://github.com/kubermatic/kubermatic/releases/latest -o /dev/null | sed -e 's|.*/v||')
+wget https://github.com/kubermatic/kubermatic/releases/download/v${VERSION}/kubermatic-ce-v${VERSION}-linux-amd64.tar.gz
+tar -xzvf kubermatic-ce-v${VERSION}-linux-amd64.tar.gz
+```
+You can find more information regarding the download instructions in the [CE installation guide](../install-kkp-ce/#download-the-installer).
+
+**2. Run the `local` command.**
+
+```bash
+./kubermatic-installer local kind
+```
+
+The KKP installation should take no longer than a couple of minutes.
+First, it creates a `kind` cluster named `kind-kkp-cluster`, exposes ports `443`, `80` for the KKP API, KKP dashboard, dex, and ports `6443`, `8088` for user cluster `kube-apiserver` via the [tunneling expose strategy](../../tutorials-howtos/networking/expose-strategies/#tunneling). You can examine the generated `kind` config, as well as the generated Kubermatic and Helm values configuration, under your working directory:
+
+```none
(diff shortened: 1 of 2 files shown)


wozniakjan:
/hold
should go to v2.23 and not v2.22

content/kubermatic/v2.22/installation/local-kkp-installation/_index.en.md
```diff
@@ -0,0 +1,83 @@
...
+```
+
+## Configuration and Customization
+
+* The KKP dashboard is [exposed using `nip.io`](https://nip.io/), and certain browser plugins, such as various ad blockers, may prevent access to `nip.io`. If this is the case, consider disabling these plugins.
+* By default, KubeVirt is configured to use hardware virtualization. If this is not possible for your setup, consider [setting KubeVirt to use software emulation mode](https://github.com/kubevirt/kubevirt/blob/v1.0.0-rc.0/docs/software-emulation.md).
+* On Linux, KubeVirt uses the inode notify kernel subsystem `inotify` to watch for changes in certain files. You may need to set the default values higher to ensure KubeVirt operates correctly.
+```bash
+sudo sysctl -w fs.inotify.max_user_watches=524288
+sudo sysctl -w fs.inotify.max_user_instances=256
```
wozniakjan, line 82: @xmudrii I used the values you provided, I tried to search KubeVirt docs for any recommendations but I couldn't find any.

Maybe @mfranczy has some experience with this?

content/kubermatic/v2.22/installation/local-kkp-installation/_index.en.md
```diff
@@ -0,0 +1,83 @@
++++
```
embik, line 1: This is the wrong place for the docs since the feature doesn't exist in v2.22. It should be moved to `content/kubermatic/main` so I can sync it from there into my PR over at #1482.

content/kubermatic/v2.22/installation/local-kkp-installation/_index.en.md
```diff
@@ -0,0 +1,83 @@
...
+## Pre-Installation Requirements
+
+- **Operating System:** Currently, only Linux is supported.
+- **Disk Space:** A minimum of 10Gi is recommended.
+- **RAM:** At least 8Gi of RAM is needed; 16Gi is recommended for more than a single KubeVirt user cluster.
+- **`kind`** (version 0.17 or higher): Please refer to the [`kind` documentation](https://kind.sigs.k8s.io/docs/user/quick-start/) for installation instructions.
+
+{{% notice warning %}}
+This command is experimental and not intended for production setups.
+{{% /notice %}}
```
embik, line 20: Please move this notice up to the top, it should be the first thing on the page. I think it would also be helpful to clarify the description, because it almost sounds like once this is no longer experimental, it _will_ be ready for production setups.

content/kubermatic/v2.22/installation/local-kkp-installation/_index.en.md
```diff
@@ -0,0 +1,83 @@
...
+
+## Installation Procedure
+
+Follow these steps to use the KKP installer local command:
+
+1. Download the installer . This is the only manual step; there's no need to prepare any configuration since the installer should automatically configure KKP.
```
embik, line 26:
```suggestion
1. Download the installer. This is the only manual step; there's no need to prepare any configuration since the installer should automatically configure KKP.
```

embik (changes requested):

content/kubermatic/v2.22/installation/local-kkp-installation/_index.en.md
embik: Sorry to come back and still complain, but I do think the page should mention that the feature is experimental. Maybe it should be two separate warnings at the top of the page though.

content/kubermatic/v2.22/installation/local-kkp-installation/_index.en.md
mfranczy: Those values are very biased and dependent on processes that watch for files on your system and if your hardware is very memory constrained (like 4Gi) it might be a bad idea to go that high.

Usually you don't need to fine-tune this just for KubeVirt, especially when you start a development setup which creates a few VMs. Also as I remember inotify was only used to trigger gracefule shutdown of VMIs and it's a legacy support https://github.com/kubevirt/kubevirt/blob/292f07f9833011246f9aa9d6780eba094d5cfd0a/cmd/virt-handler/virt-handler.go#L303-L309

I am pretty sure that KubeVirt doesn't need to go with such high values. Rather, if user has a problem I would advise that they judge how much they need to increase.

@wozniakjan why do you advise to change this? Do you have evidence that this is really required for KubeVirt? Maybe you discovered some intofy leak.

mfranczy: If you want to come with real numbers for KubeVirt I would suggest to run dev setup and check how many inotify watches are open. You can check this with `lsof`  and filter `inotify` over name column.

content/kubermatic/v2.22/installation/local-kkp-installation/_index.en.md
wozniakjan: > Sorry to come back

no worries, sometimes multiple iterations of incremental improvement is the only way to reach a consensus :)

However, two "warnings" on the top looked imho a bit awkward so I kept the most important disclaimer about "no prod" on the top and moved the second disclaimer with "experimental" a bit lower and provided more info about why it's experimental and what to expect from the resource consumption point of view. The second warning is still before installation instructions.

content/kubermatic/v2.22/installation/local-kkp-installation/_index.en.md
wozniakjan: > why do you advise to change this? 

This was discovered by @xmudrii, the `kubevirt/virt-handler` was crashing for him with `"Failed to create an inotify watcher"`, more here - https://github.com/kubermatic/kubermatic/issues/12340#issuecomment-1602433880

But the verbiage is purposefully crafted as a "you may need .. if KubeVirt is failing". You sense correctly that I would prefer not to direct users or even try to manage this inside of the installer.

> You can check this with `lsof` and filter `inotify` over name column.

This is really handy, thanks!

wozniakjan: I added additional sentence supporting ^^
```
Reasonably elevated values are described below but it's recommended to inspect your system to figure out the correct `inotify` values depending on your needs and current setup
```

embik:
@wozniakjan sorry for the back and forth, but we merged the v2.23 content folder now. Please make sure that you provide this both in `main` and `v2.23`.

wozniakjan:
@embik all good, rebased and duplicated to `v2.23` dir, ptal

content/kubermatic/v2.22/installation/local-kkp-installation/_index.en.md
embik: @mfranczy are you happy with the changes done here?

embik (approved):
/approve

Just so we have something online to explain this feature. We can adjust it later on.

PR #1488, opened 2023-06-28
Fix 'getting kubeone' zsh command completion

SamuAlfageme:
Prevents copy-paste errors, since sourcing the bash completion script on the .zshrc will generate some error messages.

### DIFF
--- a/content/kubeone/main/getting-kubeone/_index.en.md
+++ b/content/kubeone/main/getting-kubeone/_index.en.md
@@ -141,7 +141,7 @@ source <(kubeone completion zsh)
 To active completion permanently, add the command to your `~/.zshrc` file.
 
 ```
-echo "source <(kubeone completion bash)" >> ~/.zshrc
+echo "source <(kubeone completion zsh)" >> ~/.zshrc
 ```
 
 ### Generating documentation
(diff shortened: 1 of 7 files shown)


embik:
Thank you for finding this mistake @SamuAlfageme. Can I ask you to update `content/kubeone/main/getting-kubeone/_index.en.md` as well so this doesn't show again in future documentation we generate from that `main` directory?

SamuAlfageme:
@embik oops! I was redirected to the one in `kubeone/v1.6` from the `Edit this page` on https://docs.kubermatic.com/kubeone/v1.6/getting-kubeone/ but indeed, I can fix the typo for all versions of the docs as well.

embik:
> @embik oops! I was redirected to the one in `kubeone/v1.6` from the `Edit this page` on https://docs.kubermatic.com/kubeone/v1.6/getting-kubeone/ but indeed, I can fix the typo for all versions of the docs as well.

I mean, the button was right, so don't worry - our documentation is just structured this way so if the fix is something not specific to the version, we might need to ask for this. I appreciate you taking the time and going back for all releases!

embik (approved):
Thank you!

/approve

PR #1494, opened 2023-07-12
Redirect etcd backup information to single page

embik:
We have two separate pages detailing the (new-ish) etcd backup and restore functionality. The page under "Cheat Sheets" is severely outdated and I find it's in an unexpected location.

I have moved content from it into the "Tutorials & How-Tos > Automatic etcd backup and restore" page we _also_ have and which I find much better. This PR adds a redirect to redirect the previous page to the new one, so people looking for etcd backup information with a bookmark or with the knowledge that this page exists / existed will be redirected to the canonical source of information.


csengerszabo (approved):
/lgtm

PR #1499, opened 2023-07-26
Add MacOS download snippet in installation docs

cnvergence:


content/kubermatic/main/installation/local-kkp-installation/_index.en.md
```diff
@@ -29,11 +29,26 @@ Follow these steps to use the KKP installer local command:
...
 tar -xzvf kubermatic-ce-v${VERSION}-linux-amd64.tar.gz
 ```
+{{% /tab %}}
+{{% tab name="MacOS" %}}
+```bash
+# For latest version:
+VERSION=$(curl -w '%{url_effective}' -I -L -s -S https://github.com/kubermatic/kubermatic/releases/latest -o /dev/null | sed -e 's|.*/v||')
+# For specific version set it explicitly:
+# VERSION=2.21.x
+wget https://github.com/kubermatic/kubermatic/releases/download/v${VERSION}/kubermatic-ce-v${VERSION}-darwin-amd64.tar.gz
```
embik, line 46: nit: you probably want an `ARCH` variable defined before that gets the machine architecture, since macOS systems can either be AMD64 or ARM64.

embik (changes requested):

content/kubermatic/main/installation/local-kkp-installation/_index.en.md
cnvergence: Good idea, I also thought about it when applying the same to the other installation [docs](https://github.com/kubermatic/docs/pull/1449)
Let me then extend this PR

content/kubermatic/main/installation/install-kkp-CE/_index.en.md
```diff
@@ -95,12 +95,15 @@ tar -xzvf kubermatic-ce-v${VERSION}-linux-amd64.tar.gz
 {{% /tab %}}
 {{% tab name="MacOS" %}}
 ```bash
+# Determine your macOS processor architecture type
+# Replace 'amd64' with 'arm64' if using an Apple Silicon (M1) Mac.
+export ARCH=amd64
```
embik, line 100: should use `uname -m` (or something similar; I _think_ `uname -m` is the way to go but please double-check) to auto-determine the cpu type - and no need to export it.

embik (changes requested):

content/kubermatic/main/installation/install-kkp-CE/_index.en.md
cnvergence: no, you can't use it for `amd64`, it's either `x86_64` or `i386`

embik: ah damn. okay, that's unfortunate.

embik (approved):
/approve

PR #1501, opened 2023-08-02
change machine_controller_machines_total to new kube_node_info metric

WeirdMachine:
The metric based for nodes in a cluster where changed a while ago from machine_controller_machines_total to kube_node_info metric

### DIFF
--- a/content/kubermatic/main/tutorials-howtos/metering/_index.en.md
+++ b/content/kubermatic/main/tutorials-howtos/metering/_index.en.md
@@ -167,8 +167,7 @@ Kubelet Metrics:
 - machine_cpu_cores
 - machine_memory_bytes
 
-Other Metrics:
-
+Kube-State Metrics:
 - kube_node_info
 
 The following values will be written to the reports:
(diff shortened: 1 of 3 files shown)


content/kubermatic/main/tutorials-howtos/metering/_index.en.md
```diff
@@ -167,9 +167,9 @@ Kubelet Metrics:
 - machine_cpu_cores
 - machine_memory_bytes
 
-Machine controller Metrics:
+Other Metrics:
```
embik, line 170: I don't think this should be called "other metrics", since it's metrics coming from kube-state-metrics. Can we specify that somewhere? It seems that kube-state-metrics is now a requirement for metering, but the page doesn't even mention it as pre-requisite (it's an optional addon not installed into all clusters).

embik (changes requested):

content/kubermatic/main/tutorials-howtos/metering/_index.en.md
WeirdMachine: yeah I changed the titel. Turns our kube-state-metrics is deployed by our monitoring controller by default:
https://github.com/kubermatic/kubermatic/blob/main/pkg/controller/seed-controller-manager/monitoring/resources.go#L89

embik (approved):
/approve

PR #1512, opened 2023-09-02
osm: replace screenshot with the custom OSP's diff as snippet

SamuAlfageme:
The text on this screenshot is a bit illegible and replacing it with a diff snippet makes it easier to understand:

Screenshot 2023-09-02 at 17 27 55

The custom OSP that is included in `content/` was missing the `NodeSwap=true` included on the feature gates, so I included it. However, the current [master version for this OSP](https://github.com/kubermatic/operating-system-manager/blob/main/deploy/osps/default/osp-ubuntu.yaml) is `v.1.3.1` and has changed a bit ever since. Maybe it can also be updated for the example.


ahmedwaleedmalik (approved):
/approve

PR #1513, opened 2023-09-04
Add storage requirements page that mentions etcd requirements

embik:
This PR is a small addition to our requirements section that adds some notes around storage requirements. Previously, we have not mentioned this at all, but especially in private datacenters poor storage performance will make etcd very unstable.


archups (approved):
/lgtm

PR #1516, opened 2023-09-08
fix KKP addon configuration snippets

dakraus:


embik:
Hey @dakraus this looks good but the Hugo build is really brittle unfortunately. Can you please push another commit to re-trigger it?

PR #1519, opened 2023-09-18
WIP : Updated KKP CRDs documentation for few missing description

archups:
Missing CRDs description updated under ClusterSpec, ClusterStatus, OIDCSettings


embik:
This file is fully generated from our CRDs, so updating it manually will be overridden by the next sync. The sync happens in [hack/update-docs.sh](https://github.com/kubermatic/kubermatic/blob/main/hack/ci/update-docs.sh) on every commit to kubermatic/kubermatic.

Since you are updating CRD descriptions, maybe we can sit down together and I can show you where in the code you can add those descriptions?

archups:
> This file is fully generated from our CRDs, so updating it manually will be overridden by the next sync. The sync happens in [hack/update-docs.sh](https://github.com/kubermatic/kubermatic/blob/main/hack/ci/update-docs.sh) on every commit to kubermatic/kubermatic.
> 
> Since you are updating CRD descriptions, maybe we can sit down together and I can show you where in the code you can add those descriptions?

Sure let's do it!

archups:
Since the CRDs doc is auto generated via code, closing this PR. Changes would be taken care as part of the code with the new PR.

PR #1522, opened 2023-09-20
Correct support matrix for KubeOne 1.6

embik:
We were notified that the support matrix for KubeOne said that 1.6 supports Kubernetes 1.27. That's not the case and probably a copy-paste mistake in #1492.


xmudrii (approved):
Thank you!
/lgtm
/approve

PR #1533, opened 2023-10-18
Update Kubernetes support matrix for 2.24

embik:
We are removing 1.25 support in KKP 2.24 and haven't updated the support matrix in a while (missing 1.28 support, for example).

I'm also removing the note that we are dropping Kubernetes minor releases that went EOL in KKP patch releases. This has not been a policy of sig-release for a while, so let's remove it from public communication.


csengerszabo (approved):
/lgtm

PR #1534, opened 2023-10-18
Remove auth.caBundle reference

embik:
Our documentation references `auth.caBundle`. I have never seen this field, our API no longer has it, I believe this is ancient history and needs to be removed. CA bundles are configured globally since KKP 2.17.


xrstf (approved):
/approve

PR #1535, opened 2023-10-18
Add `kind: ConfigMap` in snippet for custom CA configuration

embik:
When configuring a custom CA bundle, you need to set `kind: ConfigMap` in `spec.caBundle`. Otherwise, KKP complains about the missing kind.


xrstf (approved):
/approve

PR #1536, opened 2023-10-19
Update documentation for disabling cert-manager integration

embik:
Fixes https://github.com/kubermatic/kubermatic/issues/12674.

@wozniakjan verified that `spec.ingress.certificateIssuer: {}` works to disable cert-manager integration in KKP.

This also adds a note on the `kind` for certificate issuers, pointing out you need to set it to either `ClusterIssuer` or `Issuer`.

In addition, information on extra args for the nginx-ingress-controller for the v2.23 docs was outdated. Updated that as well.


wozniakjan (approved):
/lgtm
/approve

wozniakjan:
hey @embik, lgtm and I also tried with 2.22 so we can update the docs there too
https://github.com/kubermatic/kubermatic/issues/12674#issuecomment-1770480111

PR #1545, opened 2023-10-20
bump Hugo to 0.119.0

xrstf:
I released a new kubermatic/hugo image and now we can have a consistent version everywhere (0.119.0). Maybe this will also finally fix the concurrent map access panic that sometimes happens.


embik (approved):
/approve

PR #1554, opened 2023-10-30
Update documentation to disable cert-manager integration

wozniakjan:
followup to htps://github.com/kubermatic/docs/pull/1536, I made an error when validating the instructions for v2.23.x, this is now operational and confirmed by the user https://github.com/kubermatic/kubermatic/issues/12674#issuecomment-1780088061.


ahmedwaleedmalik (commented):
Don't we need the same changes in main, v2.22 docs as well?

wozniakjan:
> Don't we need the same changes in main, v2.22 docs as well?

good point, it's probably best to change for 2.22 too to keep consistent (in 2.22 the `{}` actually works somehow...)

embik (approved):
/approve

PR #1555, opened 2023-10-30
fix AWS Permissions for KKP > 1.22

toschneck:
Based on the used role at our demo environment, the documented AWS role was missing the  permission `TagInstanceProfile`.

### DIFF
--- a/content/kubermatic/main/architecture/supported-providers/aws/_index.en.md
+++ b/content/kubermatic/main/architecture/supported-providers/aws/_index.en.md
@@ -60,6 +60,7 @@ Ensure that the assigned policy contains at least the following permissions. Pol
             "Effect": "Allow",
             "Action": [
                 "ec2:*",
+                "iam:TagInstanceProfile",
                 "elasticloadbalancing:CreateListener",
                 "elasticloadbalancing:CreateRule",
                 "elasticloadbalancing:CreateTargetGroup",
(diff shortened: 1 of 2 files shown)


content/kubermatic/main/architecture/supported-providers/aws/_index.en.md
```diff
@@ -60,6 +60,7 @@ Ensure that the assigned policy contains at least the following permissions. Pol
             "Effect": "Allow",
             "Action": [
                 "ec2:*",
+                "iam:TagInstanceProfile",
```
embik, line 63: Looking at the permission in question, I think this should only be given for `"Resource": "arn:aws:iam::YOUR_ACCOUNT_ID:instance-profile/*"`, i.e. the first block of permissions. So I would move this to line 30.

embik (changes requested):

content/kubermatic/main/architecture/supported-providers/aws/_index.en.md
toschneck: hmm I would need to test this (will do after kubecon)

xrstf: Did this ever happen? 😁

embik:
@archups quick question, did you check/test those permissions?

content/kubermatic/main/architecture/supported-providers/aws/_index.en.md
toschneck: changed it and tested it finally now :-P So it wokrs

PR #1560, opened 2023-11-07
explain custom metering reports

xrstf:
This gives some guidance on how to create a custom metering report, if need be.

I tried this Job out, worked just fine.


embik (approved):
/approve

PR #1564, opened 2023-11-07
Fix KKP filenames

xrstf:
At some point we have started to allow basically _any_ characters in filenames, which lead to a great inconsistency and even some really creative names like `[Experimental]-OPA-Mutation`.

Having consistent file and directory names is better, makes URL break less often and reduces the risk of URLs breaking in transit. To that end, this PR updates `kubermatic/main` and `kubeone/main` to only use `[a-z0-9-]` for every name, including images. Older releases were not changed so that URLs do not break (and because my OCD has its limits :grin:)

I also added a new Prowjob that verifies that we never ever get inconsistent names ever again. Thankfully OSM and KubeLB are already sane, so including them in these new checks was trivial.

And because I was at it, I also piped all kubermatic/main PNG images through `optipng` to shrink them down (without losing quality). This brought the size down from 69 to 44 MiB. Nice.


embik (approved):
/approve

PR #1565, opened 2023-11-08
Upsync 2.23 upgrade guide changes to main and v2.24

embik:
The v2.23 upgrade guide was updated in `content/v2.23` but not `content/main`, so this PR upsyncs changes to `main` and `v2.24`. In addition, it removes the "post-upgrade considerations" in the 2.24 upgrade guide since they are TBD and we are not aware of any for the moment.


xrstf (approved):
/approve

PR #1566, opened 2023-11-08
Sort components list alphabetically

embik:
The components list generated by the `hack/prepare-docs.sh` script is not sorted, which it is for previous (e.g. the 2.23 component list) ones.


xrstf (approved):
/approve

PR #1568, opened 2023-11-08
Set max-height on KubeLB logo

embik:
The KubeLB logo is a bit ... large on bigger screen. This adds a max-height helper class so it doesn't grow past a certain point, much more in line with logos on the other landing pages.

Screenshot 2023-11-08 at 14 26 05


xrstf (approved):
/approve

PR #1569, opened 2023-11-08
Update aws.en.md

hc2p:
Add missing "elasticloadbalancing:AddTags" permission.

### DIFF
--- a/content/kubermatic/main/architecture/supported-providers/aws/_index.en.md
+++ b/content/kubermatic/main/architecture/supported-providers/aws/_index.en.md
@@ -60,6 +60,7 @@ Ensure that the assigned policy contains at least the following permissions. Pol
             "Effect": "Allow",
             "Action": [
                 "ec2:*",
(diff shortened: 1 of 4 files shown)


xrstf:
How does this relate to #1555?

hc2p:
I was not aware of #1555 until now. 
This additional permission here was needed to provision AWS ELBs from a KKP 2.23.6 installation.

embik:
Can you please add this to `main` and `v2.24` for KKP docs as well?

xrstf:
(Note that the file is _index.en.md in those newer docs versions.)

embik (approved):
/approve

xrstf:
@archups Thank you so, so much for this. Can we now close #1555 as well?

archups:
> @archups Thank you so, so much for this. Can we now close #1555 as well?

#1555 is addressing different issue fix. So IMO, update to the #1555 is required instead of closing it.

PR #1572, opened 2023-11-14
Add KKP 2.24 Release Notes

xrstf:
This also adds the release notes for today's patch releases and it adds a basically blank page to main, so that for 2.25 I won't have to grep and guess where to put the changelog.


embik (approved):
/approve

PR #1578, opened 2023-11-23
Remove mention of installer during MinIO manual migration

embik:
Re-running the installer as part of this migration doesn't work, because it will fail with

```
time="2023-11-23T11:36:40Z" level=info msg="🚦 Validating existing installation…"
time="2023-11-23T11:36:40Z" level=error msg="⛔ Cannot proceed with the installation:"
time="2023-11-23T11:36:40Z" level=error msg="   expected exactly 1 Minio Pod, but found 0; cannot exec and check PVC contents"
time="2023-11-23T11:36:40Z" level=error msg="❌ Operation failed: preflight checks have failed."
```


xrstf (approved):
/approve

PR #1579, opened 2023-11-24
vsphere permissions documentation fix

dharapvj:
Based on reports and observations at multiple vsphere based KKP installation, it is clear that newly added kkp-tags related permissions were conflicting with existing permissions. We needed to get them merged properly.

This docs Pr moves those permissions in appropriate propagated and non-propagated roles so that the roles can be applied properly at vcenter UI.


hc2p (approved):

embik (approved):
/approve

PR #1580, opened 2023-11-24
kubeone manual recovery instructions updated to use ctr instead of docker

dharapvj:
- Since new CP machines do not have docker installed, the recovery instructions are updated to use available `ctr` command instead.
- also kubeone install is deprecated so updated to use kubeone apply instead.


xmudrii (commented):
Thank you for fixing this, @dharapvj!
/lgtm
/approve

PR #1582, opened 2023-11-29
Cleanup of redundant section - Terminology from installation docs

archups:
Removing redundant Terminology section as the terms are listed in the [Architecture](https://docs.kubermatic.com/kubermatic/main/architecture/)  and [Installation](https://docs.kubermatic.com/kubermatic/main/installation/) documentations and reference to those documentation are already provided wherever needed. 
To make the master/seed installation docs more precise. :)

### DIFF
--- a/content/kubermatic/main/installation/install-kkp-ce/_index.en.md
+++ b/content/kubermatic/main/installation/install-kkp-ce/_index.en.md
@@ -14,16 +14,6 @@ Typically the setup of KKP cluster, including creation of a base Kubernetes clus
 
 Expected skills and knowledge for the installation: moderate level of familiarity with cloud services (like AWS, Azure, GCP ore others) and familiarity with container and Kubernetes technologies, constructs, and configurations.
 
-## Terminology
-
-In this chapter, you will find the following KKP-specific terms:
-
-* **Master Cluster** -- A Kubernetes cluster which is responsible for storing central information about users, projects and SSH keys. It hosts the KKP master components and might also act as a seed cluster.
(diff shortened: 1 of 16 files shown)


embik:
This was done intentionally to make pages "self-contained", as new users might click through and miss the definitions on a parent page. I don't think de-duplicating it again is a good idea.

archups:
> This was done intentionally to make pages "self-contained", as new users might click through and miss the definitions on a parent page. I don't think de-duplicating it again is a good idea.

Thank you @embik  for the giving context on why we have #Terminology section on each of the installation page. With you on it, make sense that new user might miss out on the parent installation page. 
However, we are giving reference to the KKP architecture documentation.  For example like this one, on the Install Community Edition (which is guide for KKP Master cluster setup):

> If you are looking for more general information on KKP, we recommend our [documentation start page](https://docs.kubermatic.com/kubermatic/v2.24/) and the [Architecture section of the documentation](https://docs.kubermatic.com/kubermatic/v2.24/architecture/) to get familiar with KKP’s core concepts.

And if the new user is moving forward with Seed cluster setup, IMO we assume he/she should by then familiar with the concepts/terms used or can go back to the architecture documentation to get familiar with it before moving ahead. 

Hence, this is my proposal PR :)

embik:
My personal experience in conversation has been that people don't want to click "further reading" links in our documentation but prefer for the page that they are on to provide a coherent and complete unit. Especially given that those pages are the pages that people will first navigate to when they "just want to get started", I don't think it's a good idea to deduplicate this section and assume that people will have read everything we asked them to read.

embik:
We decided to keep this as is.

PR #1586, opened 2023-12-08
Use a single clusterwide Cilium NetworkPolicy as workaround

embik:
This is an improvement over the existing documented workaround, discovered by @hc2p. Requires only a single, global object.


hc2p (approved):

embik:
xref https://github.com/kubermatic/kubermatic/issues/12874

cnvergence (approved):

PR #1588, opened 2023-12-11
Mention Cilium issue on Seed installation docs

embik:
We need to increase visibility for the Cilium NetworkPolicy issue (https://github.com/kubermatic/kubermatic/issues/12874), so this adds a warning to the seed setup docs.

Also corrects some spelling to be more consistent and removes mention of CRDs being installed by kubermatic-installer (it's now done when adding a cluster as seed).


wozniakjan (approved):
/approve

PR #1589, opened 2023-12-13
Merge sig-virtualization to sig-cluster-management

wozniakjan:
As discussed in the sig meetings, sig-virtualization will merge under the sig-cluster-management


embik (approved):
/approve

PR #1591, opened 2023-12-14
Use correct example syntax for backups-restic addon in KubeOne

embik:
As per https://github.com/kubermatic/kubeone/pull/2981, the `backups-restic` addon was incorrectly documented. This updates the existing documentation as per the current behaviour, as changing it would be a breaking change.


xmudrii (approved):
/lgtm
/approve

PR #1597, opened 2024-01-08
Document Kubevirt incompatibility with Cilium

xmudrii:
Cilium doesn't support specifying Pod's MAC address and therefore it can't be used for the KubeVirt infrastructure cluster. KubeVirt infrastructure clusters that use Cilium are not able to start VMs created by KKP/machine-controller.

Upstream reference: https://github.com/cilium/cilium/issues/22119

/assign @moadqassem @embik 
cc @toschneck

### DIFF
--- a/content/kubermatic/main/architecture/supported-providers/kubevirt/_index.en.md
+++ b/content/kubermatic/main/architecture/supported-providers/kubevirt/_index.en.md
@@ -31,7 +31,11 @@ The setup has been successfully tested with:
 * CRI: containerd
 * CNI: Canal
 
-Other CRIs and CNIs should work too. However, they were not tested, so it is possible to discover issues.
+Other CRIs and CNIs might work too. However, they were not tested, so it is possible to discover issues.
+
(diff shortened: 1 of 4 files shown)


xmudrii:
/hold
@moadqassem will confirm if it's possible to workaround this with the strict mode

xmudrii:
/remove-lifecycle stale
@moadqassem Do you have any update on this issue?

moadqassem:
> Looks like this will be part of the Cilium v1.16.0 release: https://github.com/cilium/cilium/releases/tag/v1.16.0-pre.0 (https://github.com/cilium/cilium/pull/29360)

I haven't tried it yet. But seems that it should work. Last problem that I encountered with cilium was due to the fact of MAC address propagation.

xmudrii:
https://github.com/cilium/cilium/issues/22119 is resolved, so Cilium might work well with Kubevirt, but I don't have a way to confirm this. Given this, I'm going to close this PR as we shouldn't document something we're not sure about.
/close

PR #1603, opened 2024-01-16
Update manual etcd restore procedure

embik:
Our docs were incorrect when trying to restore from legacy backups but already using etcd-launcher (since we migrated everyone in KKP v2.22 to it, essentially). This came back to bite me in the manual cluster restore I had to do on the captain.

This PR updates the manual restore procedure accordingly, but also puts a word of warning on it - no one should have to do manual restores anymore, it's all automated with the "new" backup controllers.


xrstf (approved):
/approve

PR #1609, opened 2024-01-26
document certificate renewal within kubeone

SimonTheLeg:
Hi, as this came up a two times in support tickets, I wanted to document this.

The ideas behind my wording were the following:
* I wanted to document the 90days period your certificates need to be within, before they are renewed, since I could not find that in the docs
* I wanted to document, that there is no need for a kubeadm command, but that kubeone can do it directly as long as you pass the `--force-upgrade` flag
* I wanted to give the user a clear warning that this should not be needed if they update their clusters regularly.

I am hoping these points come across in what I have written. ptal.

I am also happy to backport this to other kubeone versions of the documentation as well, just want to hear your opinion on the wording first before doing so.

### DIFF
--- a/content/kubeone/main/guides/certificate-management/_index.en.md
+++ b/content/kubeone/main/guides/certificate-management/_index.en.md
@@ -1,0 +1,18 @@
++++
+title = "Certificate Management"
+date = 2024-01-10T15:50:00+02:00
++++
+
+## Certificate Renewal
+
+{{% notice warning %}}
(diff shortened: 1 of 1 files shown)


ahmedwaleedmalik (approved):
/approve

ahmedwaleedmalik:
Once https://github.com/kubermatic/docs/pull/1615 is merged this will be good to go as well.

PR #1612, opened 2024-02-05
Update support matrix for KKP 2.25 and Kubernetes 1.29

embik:
This updates our Kubernetes support matrix according to what will be in KKP 2.25.


xrstf (approved):
/approve

PR #1624, opened 2024-03-05
fix app name: cert-manager is the official, real name

xrstf:
See https://cert-manager.io/docs/faq/#how-to-write-cert-manager

> cert-manager should always be written in lowercase. Even when it would normally be capitalized such as in titles or at the start of sentences. A hyphen should always be used between the words, don't replace it with a space and don't remove it.


ahmedwaleedmalik (approved):
/approve

PR #1630, opened 2024-03-06
rip out kubecarrier docs, gently without removing them

xrstf:
As per Sebastian's request, we remove the KubeCarrier docs now finally.


ahmedwaleedmalik (approved):
/approve

PR #1632, opened 2024-03-07
Update example base domain url to kkp.example.com

archups:
Related to https://github.com/kubermatic/kubermatic/pull/13159


xrstf (approved):
/approve

PR #1635, opened 2024-03-11
Add Flatcar to supported OSes for GCP and VCD

embik:
Ref: https://github.com/kubermatic/kubermatic/pull/13162.

It seems we haven't updated our support matrix, but we support Flatcar in GCP and VCD now.

Also does a minor update to language about Hetzner's CentOS 7 support since the date was in the past.


ahmedwaleedmalik (approved):
/approve

PR #1639, opened 2024-03-11
Minor updates to the KKP docs

archups:
Minor fixes to the installation and application catalog


content/kubermatic/main/installation/install-kkp-ee/_index.en.md
```diff
@@ -57,6 +57,17 @@ spec:
...
+```yaml
+kubermaticOperator:
+  # insert the Docker authentication JSON provided by Kubermatic here
+  imagePullSecret: |
+    {
+      "auths": {
+        "quay.io": {....}
+      }
+    }
+```
```
embik, line 69: I don't think this is necessary, because the installer will fill this in from the passed `KubermaticConfiguration`. At least all our EE setups don't need this in their `values.yaml`.

embik (changes requested):

content/kubermatic/main/installation/install-kkp-ee/_index.en.md
archups: At the moment, we have it [here](https://github.com/kubermatic/kubermatic/blob/main/charts/values.example.ee.yaml#L75) included.

embik (approved):
/approve

PR #1644, opened 2024-03-12
fix: ID of the cluster is required instead of the name

Skarlso:


Skarlso:
Extra info: When I was doing the training and testing things out, if I provided the name of my project, which was `KKP Test` it failed with not finding things. If I provided the ID of it `kkp-test`, everything worked fine.

PR #1651, opened 2024-03-13
Fix leftover from copy-paste (2.24 instead of 2.25) in upgrade guide

embik:
What it says on the title.

/cc @xrstf


xrstf (approved):
/approve

PR #1663, opened 2024-04-11
Use v2.25 images in v2.25 cluster backup documentation

embik:
Looks like the user cluster backup documentation was using `main` images (my bad, I missed this in the PR review for those docs) for the v2.25 version of the page, so they broke, likely due to #1654.

This brings back the images on this page.


xrstf (approved):
/approve

PR #1666, opened 2024-04-18
Add release notes for KKP Apr 2024 patch releases

archups:
This PR adds release notes for KKP Apr 2024 patch releases


embik (approved):
/approve

PR #1667, opened 2024-04-19
Refresh CRD documentation for minor releases

embik:
Our CRD documentation for release branches is out of date with the new patch releases because we don't automatically update it on release. Maybe we should.

Be that as it may, this PR refreshes them for all release branches. I've also made the script a bit more flexible by adding the `OUTPUT_VERSION` environment variable.


archups (approved):
/lgtm
/approve

xrstf (approved):
/approve

PR #1668, opened 2024-04-23
improve feature comparison for KKP

xrstf:
I have asked myself too often whether something is in CE or not. Others have, too. The common thing to have is a feature matrix, where I can compare the editions of a product.

This PR refactors and extends (but far from completes!) the KKP edition page.


csengerszabo (changes requested):
We could also change the "How to use Enterprise Edition" section to the suggested one by @xrstf

PR #1676, opened 2024-05-02
update Master/Seed MLA installation documentation - kubermatic-installer

mohamed-rafraf:
Change the Master/Seed MLA installation guide. The new guide explains how to install Master/Seed MLA stack with kubermatic-installer instead of using helm binary.


embik (approved):
/approve

PR #1708, opened 2024-07-09
Update docs related to services catalog

Waseem826:


content/developer-platform/platform-users/consuming-services/service-catalog.jpg
embik, line 1: Can you maybe take this screenshot on the demo environment?  This one has a lot of clutter from our tests on dev, and I'd prefer to not show e.g. the "Matrix" thing on our official docs.

I also wonder if it should be cropped a little to show it's a popup, but not the whole page.

Waseem826: We don't have this latest UI on the demo environment but I get your point and I'll update the screenshot.

Waseem826: Updated!

embik: True, sorry! I didn't think that far. My bad. Thanks for getting it updated anyway!

embik (approved):
/approve

thanks!

PR #1710, opened 2024-07-16
Move the images near to the content

archups:
This PR is to move the images near to the content. related to restructuring of the images rendering https://github.com/kubermatic/docs/pull/1654
Plus some cleanup of the images that are not used/redundant or are from previous version in the `main` branch.
Plus few broken images link fixes.
Also, updated few images.

PR #1711, opened 2024-07-17
fix image URLs in KKP main branch docs

xrstf:
This supersedes #1710.

When @ URLs are used, the release name (main or v21.2.3.12) must not be part of the URL in the Markdown file. Because image URLs are by default relative to the version they are in.


embik (approved):
/approve

PR #1717, opened 2024-07-24
Fixed branch name in the document link

archups:
This PR is to fix the branch name from `master` to `main` as per the recent change to the kubermatic/docs default branch to fix the broken document links.


embik (approved):
/approve

PR #1719, opened 2024-07-25
Bump js libraries

serg:
should fix [#134](https://github.com/kubermatic/docs/security/dependabot/134)
/assign @embik


embik (approved):
/approve

PR #1720, opened 2024-07-26
Fix more images

xrstf:
This PR

* fixes all remaining broken images I could find
* restores missing images
* removes the executable bit from all Markdown and PNG files


embik (approved):
/approve

PR #1739, opened 2024-08-30
correct description for automatic node updates

dakraus:
This PR corrects the description to configure automatic node updates in the `KubermaticConfiguration` for all documented KKP versions. Looking at the initial implementation in https://github.com/kubermatic/kubermatic/pull/4258, it seems that the field was always named `automaticNodeUpdate`.


xrstf (approved):
/approve

PR #1740, opened 2024-09-10
AWS permission fix for kubeone, based on #1555

toschneck:
- tested this new permissions on AWS for a standard and private
  networking setup
- may not all permissions are only for machinecontroller required, but
  for apply kubeone with the terrafrom example (what I think will be the
99% case), so as we don't have other AWS permissions document for
kubeone, I would see the terraform required permissions in scope to
enhance the expierence of our customers.


embik (approved):
/approve

PR #1761, opened 2024-10-04
duplicate KKP docs for v2.26

xrstf:
`hack/prepare-release.sh -p kubermatic -v v2.26`

PR #1765, opened 2024-10-08
Hide 2.26 documentation from docs.kubermatic.com until the release is out

embik:
#1761 was a bit too early, we don't want to the 2.26 documentation to be public yet.


ahmedwaleedmalik (approved):
/approve

PR #1766, opened 2024-10-09
remove oauth upgrade hint from the KKP 2.26 upgrade guide

xrstf:
There is no new `dex` chart in KKP 2.26 (https://github.com/kubermatic/kubermatic/pull/13486 didn't make it in), so telling admins already to upgrade (to what then?) now is confusing and will just cause problems with every subsequent 2.26.x patch release.


embik (approved):
/approve

PR #1776, opened 2024-10-16
fix linebreaks in generated CRD reference

xrstf:
Wne updating to 0.0.12, we forgot to include the `markdownRenderFieldDoc` call, which broke the generated markdown.

There is now a new shortcode as well, so the generated <br/> will actuallly be rendered as a line break and not stripped out by Hugo.


embik (approved):
/approve

but we will have to also re-generate the 2.26 CRD ref.

PR #1779, opened 2024-10-17
fix minio bucket name in _index.en.md documentation

zwindler:
Trying to apply this manifest will fail

You can create buckets in MinIO with "/" in them, only lowercase letters, numbers dots and hyphens

I tried to reproduce on the MinIO UI, which is more verbose

image

### DIFF
--- a/content/kubermatic/v2.26/installation/install-kkp-ce/add-seed-cluster/_index.en.md
+++ b/content/kubermatic/v2.26/installation/install-kkp-ce/add-seed-cluster/_index.en.md
@@ -235,7 +235,7 @@ the `Seed` resource has been created.
 
 If you are using MinIO, a bucket needs to be created for cluster backups to be stored in. This can be done for example
 via a `Job` resource that spawns a `Pod` running the `mc` command against the freshly deployed MinIO service. Below you
-find an example `Job` definition. If you want to change the bucket name, replace `src/kkpbackup` with `src/YOUR_BUCKET_NAME`
+find an example `Job` definition. If you want to change the bucket name, replace `kkpbackup` with `YOUR_BUCKET_NAME`
 in the `args` part of the template.
 
 ```yaml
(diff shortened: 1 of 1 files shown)


embik:
Hey, thanks for the contribution! I haven't looked at why this might be failing yet, but the `mc mb` [documentation](https://min.io/docs/minio/linux/reference/minio-mc/mc-mb.html#parameters) is pretty clear on this:

> For creating a bucket on MinIO, specify the [alias](https://min.io/docs/minio/linux/reference/minio-mc/mc-alias-set.html#alias) and the name of the bucket. For example:
> ```
> mc mb play/mybucket
> ```

`src` is not part of the bucket name here, but it's the alias for the MinIO server. You don't need this when interacting with MinIO directly via its UI, of course.

Have you verified that this actually creates a bucket and not a local directory?

embik:
/hold

to clarify.

embik:
I suspect maybe that the `host add` command no longer serves as an alias, so `src` is no longer recognised as the target MinIO ...

zwindler:
Well, I tear down my test cluster but as I recall job kept failing when I tried running it on KKP 1.25.11

I should have kept the error messages but I don't think I did... I'll have to create a new cluster to try again

embik:
> Well, I tear down my test cluster but as I recall job kept failing when I tried running it when installing KKP 1.25.11
> 
> I should have kept the error messages but I don't think I did... I'll have to create a new cluster to try again

All good, we can also investigate. I don't doubt it's failing, I was just very surprised to see this as a solution because my understanding of the tool (so far) was that it would just create a local directory. I could be wrong there though!

embik:
Hm, I cannot reproduce any error unfortunately. I've changed it to `src/test-marvin-please-ignore` to make sure I'm creating a new bucket, but the log output looks fine:

```sh
$ k logs create-minio-backup-bucket-3-ct5gs
Added `src` successfully.
Bucket created successfully `src/test-marvin-please-ignore`.
```

It would be interesting to see the errors you've encountered.

zwindler:
Interesting, I'll try again. Thanks

zwindler:
Ok, I tried again and I can't reproduce. I don't know what happened nor why I though changing the command fixed it. Sorry for the time you lost on this

embik:
> Ok, I tried again and I can't reproduce. I don't know what happened nor why I though changing the command fixed it. Sorry for the time you lost on this

No worries at all! We appreciate the spirit of contribution. Perhaps MinIO wasn't up yet and that produced a confusing error message.

PR #1781, opened 2024-10-17
remove misleading dex migration note for KKP 2.26

xrstf:
This should have also been removed in #1766, but was overlooked.


embik (approved):
/approve

PR #1786, opened 2024-10-23
Adjust docs after the apiserver allowlist revert

cnvergence:
https://github.com/kubermatic/kubermatic/issues/13541#issuecomment-2416882524


xrstf (commented):
This fix also isn't in the main branch, right? Would those docs also need to be adjusted?

ahmedwaleedmalik (approved):
/approve

PR #1797, opened 2024-11-18
Add guide to use servlet without KDP and update servlet docs

embik:
Since https://github.com/kubermatic/developer-platform/pull/286, the way the KDP servlet works has been changed. This PR adjusts our KDP docs accordingly.

In addition, it adds a guide on how to use the servlet without KDP. This is mostly for our partners and customers who want to try it out, but we don't want to give away the whole of KDP yet.


Waseem826 (approved):
/approve

PR #1800, opened 2024-11-20
Add KubeOne v1.9 to products.yaml

xmudrii:
This PR adds KubeOne v1.9 to `products.yaml` which will make the v1.9 docs public.

/assign @kron4eg 
/hold
unhold before cutting the final release


kron4eg (approved):
/approve

PR #1811, opened 2025-01-22
Os compatibility update

koksay:
Updating the docs as per the changelog.

I'm not sure if this file also need an update: `architecture/supported-providers/edge/_index.en.md`

> Currently, the edge provider only support Ubuntu 20.04 and 22.04 as an operating system. More support for other operating system will be added in the future.`


ahmedwaleedmalik (approved):
/approve

PR #1835, opened 2025-03-12
KKP: Add a draft/template for KKP upgrade docs

ahmedwaleedmalik:


content/kubermatic/main/installation/upgrading/upgrade-from-2.27-to-2.xx/_index.en.md
```diff
@@ -0,0 +1,21 @@
++++
+title = "[DRAFT] Upgrading to KKP 2.xx (Development Version)"
```
archups, line 2: Since the upgrade would be from 2.27. It is safe to replace the 2.xx with 2.28 which would be our next minor release. Replace all the occurrence from 2.xx with 2.28 in this doc. 

Also IMO safe to remove the prefix [DRAFT] from the title, since it is `main` branch and understanding is that it will be incrementally updated via PRs for any upgrade topics related to the KKP minor release 2.28. Thanks!

ahmedwaleedmalik: I intentionally kept it as 2.xx since it's a template not the final document. Before the release you are supposed to duplicate this document and then add your conversions like 2.28 etc etc.

PR #1846, opened 2025-03-18
rephrase KDP docs, s/Servlet/api-syncagent/g

xrstf:
KDP uses kcp api-syncagent instead of the old servlet. This PR updates the wording.


embik (approved):
/approve

PR #1847, opened 2025-03-21
improve app documentation

soer3n:
This pr improves the documentation about usage of application templating and adds an explanation how to migrate from addons to autoscaler application.

### DIFF
--- a/content/kubermatic/main/architecture/concept/kkp-concepts/applications/application-templating/_index.en.md
+++ b/content/kubermatic/main/architecture/concept/kkp-concepts/applications/application-templating/_index.en.md
@@ -18,3 +18,36 @@ Go snippet shows the available information:
 ```
 {{< readfile "kubermatic/main/data/applicationdata.go" >}}
 ```
+
+### A practical example
+
+Let's take nginx ingress controller as an example to demonstrate how to use predefined values and also injecting helm variables for values which are wrapped by a `tpl` function inside the chart. Here is an example application definition which configures two configmap values rendered by helm. One with the cluster name passed from pre-defined values and the other one uses a helm variable to set the release namespace as a value.
+
+```yaml
+apiVersion: apps.kubermatic.k8c.io/v1
+kind: ApplicationDefinition
+metadata:
+  name: nginx-demo
+spec:
+  description: Demo app for testing pre defined values and helm variables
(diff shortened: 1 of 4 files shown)


content/kubermatic/main/architecture/concept/kkp-concepts/applications/application-templating/_index.en.md
```diff
@@ -18,3 +18,36 @@ Go snippet shows the available information:
...
+  - template:
+      source:
+        helm:
+          chartName: ingress-nginx
+          chartVersion: 4.12.0
+          url: https://kubernetes.github.io/ingress-nginx
+    version: 1.12.0
+  defaultValuesBlock: |
+      tcp:
+        namespace: '{{ `{{.Release.Namespace}}` }}'
```
SimonTheLeg, line 45: I am not sure if we want to include the special double curly bracket `tpl` func case here. I think this might confuse readers in two ways:
a) the reader might get the impression that this is something which works everywhere in helm. But it requires a super specific edge case solution where the upstream function needs to use the `tpl` func on a subset of a resource. 
b) the reader might get the impression that this is something which is not a helm native feature, but rather soemthing you can do in AppDefs

Now if you say I really want this in, then my suggestions would be as follows:

_Let's take nginx ingress controller as an example to demonstrate how to use predefined values. Additionally if the upstream chart is using [helms native `tpl` function](https://helm.sh/docs/howto/charts_tips_and_tricks/#using-the-tpl-function) (see example below) we can also use the double curly bracket syntax to inject helm context values._

And then below the last sentence, put in an example section of the upstream chart (I forgot the concrete syntax, but I think you know what I mean):

```yaml
# upstream chart
{{ tpl . }}
```

SimonTheLeg (changes requested):

content/kubermatic/main/architecture/concept/kkp-concepts/applications/application-templating/_index.en.md
soer3n: Thanks for the tip. I've streamlined the section a bit and, as you suggested, added an example section on the tpl function in Helm template files. The usage of predefined values ​​and Helm values ​​has been already requested, so i think it makes sense to explain it.

PR #1852, opened 2025-03-27
Update _index.en.md fix loki-ingester storage customization part

AndrStnz:
Current doc for Loki-distributed persistence is broken from at least v2.25 on. This is my proposed fix for the documentation. Thanks!


dharapvj (approved):
I agree.

Due to major upgrade to sync up with latest from upstream charts, we made these changes and documentation has not reflected this change.

Thank you for pointing it out.

/lgtm

dharapvj:
I think the same documentation change must be done in all the further versioned folders and in main branch folder as well.

PR #1858, opened 2025-04-23
Improving the creating-clusters-baremetal tutorial in KubeOne docs

ttuellmann:
This fixes three minor issues in the KubeOne tutorial "Creating a Kubernetes Cluster on Bare-Metal":

- The number of control plane nodes should be odd rather than even
- The manifest in step 3 used the taint `node-role.kubernetes.io/master` on the control plane nodes, which has been removed in all supported K8s releases.
- I bumped the K8s release to 1.32.3 in the manifest


ttuellmann:
@kron4eg No worries, thanks for getting it merged.

PR #1861, opened 2025-04-27
Doc for using kubectl oidc-login with KKP

koksay:
When the OIDC Kubeconfig feautre is enabled, it's sometimes frustrating to log in to KKP and download the kubeconfig for the user clusters, especially for the short living tokens (e.g. 1 day).

With kubelogin kubectl plugin, users will not need to navigate on the KKP UI to find and download kubeconfig, instead a browser window will be opened to re-authenticated every time the token is expired. We can help to have a better user experience with documentation.


judge-red:
Maybe you should lose a word on the need to add `http://localhost:8000` as a redirect URI to the `exampleIssuer` - otherwise end users will just end up with an error from Dex.

koksay:
Thanks for this feedback, I knew I forgot something!

kron4eg:
/lgtm
/approve

:tada:

PR #1881, opened 2025-06-02
Add feature staging guidelines

ahmedwaleedmalik:


content/kubermatic/v2.27/architecture/feature-stages/_index.en.md
```diff
@@ -38,3 +40,16 @@ weight = 4
 - All changes to the feature are backwards compatible with an automated migration path implemented where needed
 - Removing the feature requires a strict deprecation cycle and replacement or migration path when possible
 - Recommended for production use
+
+### Feature Stage Guidelines
+
+- The following components should explicitly be marked as alpha/beta:
+  - APIs/CRDs: Dedicated CRDs including fields in API and CRD that are introduced for this feature.
+  - UI: All the views/fields that are used to interact with the feature.
```
csengerszabo, line 48: Newly introduced features in general should rather always have a switch in the admin panel which turns them fully off and on (besides the feature gate).

Therefore, if the platform admin decides to turn a feature on, it's enough to have these marks for the platform admin, as from that point, it's their responsibility to let their custumers / users use these features. Marking features throughout all the user personas of KKP could be confusing to the end users in several cases.

The possibility to turn a feature on and off by the Platform Admin of KKP should solve this issue, and the Platform Admin should always be 100% aware if a feature is in Beta.

csengerszabo (changes requested):

PR #1883, opened 2025-06-04
Update documentation for Dex

ahmedwaleedmalik:


mohamed-rafraf:
After internal discussion with @ahmedwaleedmalik, there is no need to tell users to disable the new Dex chart since we are forcing the users to use this one. 

Thank you 
/approve
/lgtm

PR #1887, opened 2025-06-11
Docs for disable user SSH Key

ahmadhamzh:
ref: https://github.com/kubermatic/dashboard/pull/7324


ahmadhamzh:
> Should we also mention implications of this flag on the KKP dashboard UI?

isn't `When this feature flag is enabled, the User SSH Key Agent will be disabled.` should say that, 
or maybe we can also add `will be disabled across platform.` for example ? or WDYT?

PR #1920, opened 2025-07-21
Improved the language for ApplicationInstallation page

buraksekili:


buraksekili:
Thank you @archups! I've applied your suggestions

PR #1921, opened 2025-07-22
add info about kubelb tenant defaulting at kkp project level

soer3n:
This pr adds information about defaulting of kubelb tenants at kkp project level. ref: https://github.com/kubermatic/kubermatic/pull/14819


archups (changes requested):
Suggested change to the reference tenantspec link. 
Also since KKP 2.29 release is out, please replicate the content changes in kubermatic/v2.29 directory as well with tenantspec link adjusted to compatible KubeLB version (instead of pointing to `main`)
Thanks!

PR #1939, opened 2025-07-30
Setup documentation for the  KDP spec generator

julioc-p:
Instructions to set up AI-Agent for spec generation in KDP and use it


julioc-p:
/cc @embik 
I am assigning you since you are packaging everything right now to see if this is enough for your case.

PR #1949, opened 2025-08-06
add kubev installer doc

mohamed-rafraf:


moadqassem:
@mohamed-rafraf Please fix the lint errors so we can merge the doc

PR #1980, opened 2025-09-15
Bump the crd-ref-docs version to v0.2.0 and Go build image to 1.25.1

archups:
This PR is to fix the `post-kubermatic-update-docs` prow job failure. 

```
# golang.org/x/tools/internal/tokeninternal
../../../../pkg/mod/golang.org/x/tools@v0.19.0/internal/tokeninternal/tokeninternal.go:64:9: invalid array length -delta * delta (constant -256 of type int64)
```


kron4eg (approved):
/approve

PR #1982, opened 2025-09-18
Bump hugo image version to 0.150.0-0

archups:
This PR is to bump hugo image version to 0.150.0-0 from quite older version 0.119.0-0 in use.

### DIFF
--- a/.prow.yaml
+++ b/.prow.yaml
@@ -35,7 +35,7 @@ presubmits:
     clone_uri: "ssh://git@github.com/kubermatic/docs.git"
     spec:
       containers:
-        - image: quay.io/kubermatic/hugo:0.119.0-0
+        - image: quay.io/kubermatic/hugo:0.150.0-0
           command:
             - "./hack/ci/verify-hugo.sh"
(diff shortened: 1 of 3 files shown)


kron4eg (changes requested):
Please also update the netlify.toml

archups:
> I think we need to switch the base image to more modern: https://docs.netlify.com/build/configure-builds/overview/#build-image-selection

The current build image in use seems to be Ubuntu 20.04 (focal) as per `deploy/netlify` job's "Initialzing" stage log 
`12:16:31 AM: build-image version: 235a96ed6c11cbdf942eed7e70e6802956f9b464 (focal)" `

Looks like the switching of the build image to Ubuntu 22.04 is required as per the upstream issue https://github.com/gohugoio/hugo/issues/13942 to make it work.

serg:
The image was updated.
We are getting many warnings during build with 0.150.0 version.
Please check if its affects on the content presentation.

kron4eg (approved):
/approve

PR #1986, opened 2025-09-19
Fix post-kubermatic-update-docs job for KKP release 2.27 and 2.26

archups:
Previous PR https://github.com/kubermatic/docs/pull/1980 fixed the `post-kubermatic-update-docs` prow job failure for KKP  `main` and 2.28, however broke for the 2.27 and 2.26. This PR is to follow-up to fix post-kubermatic-update-docs-x.xx job failures for KKP release 2.27 and 2.26.  When Go version is <1.24, then `github.com/elastic/crd-ref-docs@v0.1.0` should be installed instead of `github.com/elastic/crd-ref-docs@v0.2.0`.

### DIFF
--- a/hack/render-crds.sh
+++ b/hack/render-crds.sh
@@ -29,9 +29,18 @@ if [[ -z "$SOURCE" ]]; then
 fi
 
 which crd-ref-docs >/dev/null || {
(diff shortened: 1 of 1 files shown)


kron4eg (approved):
/approve

PR #1992, opened 2025-09-25
refine Seed MLA docs

buraksekili:


archups:
Thank you for the PR, the doc refinement looks good to me! 
Small request to replicate the changes in the supported releases 2.28, 2.27 and 2.26 since it seems to be applicable to those as well right?

buraksekili:
@archups thanks for the review! I've updated the versions you mentioned accordingly.

PR #2066, opened 2026-01-16
Add quick start guide for KubeOne

scheeles:


content/kubeone/main/quick-start/_index.en.md
```diff
@@ -0,0 +1,18 @@
...
+title = "Quick Start"
+date = 2026-01-16T20:30:00+01:00
+weight = 1
++++
+
+The fastest way to deploy a KubeOne Kubernetes cluster is to use the installation script:
+
+```bash
+curl -sfL get.kubeone.io | sh
+kubeone local
```
kron4eg: This command is a bit dangerous (in the sense that it will turn local machine into the cluster), IMHO we should warn here about consequences.

PR #2067, opened 2026-01-16
Change Static Worker to Static Nodes

scheeles:
Reason for changing the headline, we are not only managing the worker nodes. At the end K1 is managing static CP + worker


kron4eg (commented):
we should add guides/static-workers -> guides/static-nodes redirect

PR #2143, opened 2026-04-23
docs(kubelb): replace kgateway with agentgateway for AI & MCP gateway

ahmedwaleedmalik:
## Summary

Rewrites the AI & MCP Gateway tutorial end-to-end for agentgateway and removes kgateway from the management-cluster install example and the air-gap image lists. Part of the KubeLB v1.4 doc rollup, kgateway is being dropped from `kubelb-addons` in v1.4.


Waseem826 (approved):
/approve

PR #2151, opened 2026-04-24
docs(kubelb): document Ingress backend-protocol HTTPS/GRPC behavior

ahmedwaleedmalik:


Waseem826 (approved):
/approve

ahmedwaleedmalik:
We should rather treat this as an internal detail instead of user facing change; nothing actually changes for the user.

PR #2186, opened 2026-05-22
fix bin path on using kubermatic-ee-installer, fix env vars on using …

stroebitzer:
…kubermatic-ee-installer


kron4eg (approved):
/approve

PR #2195, opened 2026-06-17
Remove unused rule

serg:
/assign @kron4eg


kron4eg:
@serg but why is it unused? I can see llms.txt is ptesent

serg:
@kron4eg this instruction doesnt follow the allowed rules for robots.txt. the robots.txt is invalid with this rule. removing wont affect on current recognition for llms.txt

kron4eg (approved):
/approve

PR #2204, opened 2026-06-30
KubeLB: fix command for SA creation of KKP integration

toschneck:
Fix command how to creat a KubeLB <> KKP connection service account


ahmedwaleedmalik (approved):
Thank you 😅 

/approve

PR #2220, opened 2026-07-13
docs(kubelb): tenant-managed WAF policies

ahmedwaleedmalik:
Documents tenant-managed WAF policies for KubeLB. Tenants author a namespaced `TenantWAFPolicy` in their own cluster, the operator enables it via `Config`/`Tenant`, and admin plus tenant policies run as chained WAF engines so a tenant can only add protection, never weaken the baseline.


Waseem826 (approved):
/approve

PR #2224, opened 2026-07-21
docs(kubev): document required top-level imagePullSecret

mihiragrawal:
Since v1.2.0 the controller-manager is always deployed and reads its registry credentials from the top-level imagePullSecret field, which also feeds the dashboard and api-server. The declarative-installation page only documented dashboard.imagePullSecret, so configs upgraded from v1.1.x leave the controller-manager in ImagePullBackOff (401).
- Add a dedicated "Image Pull Secret" section that explains the requirement once, covering both the top-level field and the KUBEV_USERNAME/KUBEV_PASSWORD environment-variable alternative.
- Reference that section from the Minimal, Complete, and Step 1 configuration examples instead of repeating the explanation.
- Remove the now-redundant dashboard-specific pull-secret examples. Applied to both the v1.2.0 and main pages.


mohamed-rafraf (approved):
/approve 
/lgtm

PR #2225, opened 2026-07-22
docs(kubelb): document Envoy client header size limits

### LINKED ISSUE #511: Option to configure maximum allowed client header sizes for Envoy.
Currently, admins cannot increase the max allowed client header sizes for Envoy, resulting in clients receiving `431 Request Header Fields Too Large` when any of the client's headers are larger than Envoy's default of 60KB. 

Admins should have an option to configure it, or larger defaults should be used.

ahmedwaleedmalik:
Adds a "Client header size limits" section to the KubeLB management-cluster install guide.

It documents `Config.spec.envoyProxy.headerLimits`, which raise Envoy's 60 KiB client header cap on KubeLB's managed Envoy so requests with large auth/JWT tokens or long cookies stop getting rejected with `431 Request Header Fields Too Large`. It also explains that the Envoy Gateway edge is not managed by KubeLB and shows how to raise its limit separately.

Goes with kubermatic/kubelb#519, which fixes kubermatic/kubelb#511.


Waseem826 (approved):
/approve

PR #2229, opened 2026-07-23
docs(kubelb): dedicated AI section for gateway setup, budgets and virtual keys

ahmedwaleedmalik:
Replaces the single `tutorials/aigateway` page with an AI chapter split by role, documenting the multi-tenant AI gateway shipping in KubeLB v1.5:

- `tutorials/ai/gateway/` - platform admin: enabling the addons and `Config.spec.ai`, provider backends and credential custody, the managed resources and policy-composition rules, direct MCP/A2A usage, air-gap.
- `tutorials/ai/budgets-and-virtual-keys/` - tenants: self-service `VirtualKey` lifecycle, budget levels and enforcement caveats, live spend via status, the client error contract, and the `kubelb:*` showback metrics contract.

The old `tutorials/aigateway/` URL is kept as an alias


Waseem826 (approved):
/approve

PR #2230, opened 2026-07-23
Add Configuration Maximums section for Kubermatic Virtualization

mihiragrawal:
This PR adds a new **Configuration Maximums** reference for Kubermatic Virtualization.
It documents the tested capacity limits of a single KubeV cluster and the factor that constrains each one.

Every figure was measured on a live reference cluster using the in-cluster ConfigMax benchmark and validated as functional rather than estimated. The reference presents the validated per-cluster maximums for virtual machines, networks, firewall policies, subnets, services, and routes; the limiting component for each capability; a tenant-density degradation result that establishes how many tenants a cluster sustains before running workloads begin to slow down; instructions for reproducing each figure on a target cluster through either the operator or the CLI and an engineering reference that documents the measurement methodology and the cluster tuning on which the results depend.


moadqassem (approved):
/approve
/lgtm

PR #2238, opened 2026-07-29
Render mermaid diagrams and add click-to-expand

abuango:
Mermaid diagrams have never rendered on the docs site. They show as plain code blocks.

**Why**
- Pages use ` ```mermaid ` fences, which Hugo renders as `<pre><code class="language-mermaid">`.
- The theme's `mermaid` shortcode is used by no page, and nothing loads `mermaid.js`. `themes/kubermatic-docs/static/mermaid/` has been orphaned since the theme restructure (#1170, Aug 2022).
- That vendored bundle is too old for the syntax in use anyway — it throws `Parse error` on `subgraph "User Browser"`.

**Changes**
- `static/mermaid/mermaid-11.16.0.min.js` — mermaid 11.16.0, vendored and pinned in the filename. No CDN, so diagrams work in air-gapped builds and the build stays plain `hugo --minify`.
- `layouts/partials/mermaid-loader.html` — loads mermaid once per page, renders every `.mermaid` block, wires the lightbox.
- `layouts/_default/_markup/render-codeblock-mermaid.html` — renders ` ```mermaid ` fences. Fences keep working on GitHub.
- `layouts/shortcodes/mermaid.html` — overrides the theme shortcode to use the same loader, so `{{< mermaid >}}` works for the first time.
- `static/js/mermaid-lightbox.js` — click a diagram to open it full-screen, with scroll-to-zoom, drag-to-pan, Fit, and Esc to close. The dense diagrams are unreadable inline.
- Deletes the four orphaned 2022 assets (`mermaid.js` plus three unused CSS files, 919 lines). Nothing referenced them.

**Affects** 8 pages: SecureGuard, KubeLB, and Developer Platform v0.9.0/v1.0/v1.1. All verified rendering locally, with no CDN requests.

**Size:** the bundle is 3.4 MB in the repo, 0.93 MB gzipped over the wire, and only requested on pages that contain a diagram. Bundling from npm via Hugo Pipes instead would need `npm ci` added to the Netlify and Prow builds, so vendoring keeps this PR to content and layouts.


ahmedwaleedmalik (approved):
/approve

PR #2239, opened 2026-07-29
kubelb: document Insights

ahmedwaleedmalik:
Documents KubeLB Insights: what the engine reports, how to triage a finding, how to suppress a check, and the posture score.

Two pages under a new `insights` chapter: the guide, and a check catalog with one section per check. Findings link into the catalog by anchor.


Waseem826 (approved):
/approve

PR #2240, opened 2026-07-29
docs(kubelb): add the mTLS rotation check to the insights catalog

ahmedwaleedmalik:
Adds the catalog entry for `mtls-rotation-stuck`, which reports backend transport certificates that entered their rotation window without being reissued, plus the MTLS row in the capability table.


Waseem826 (approved):
/approve

PR #2243, opened 2026-07-29
docs(kubelb): bump hardcoded versions on security pages to v1.4.3

ahmedwaleedmalik:
The Verifying Artifacts page hardcodes kubelb release versions in its cosign/oras/trivy examples, release download URLs and SBOM filenames, and the release docs automation never bumped it: `v1.4` tree was stuck at v1.4.1, `main` at v1.3.0.

Bumps both to v1.4.3. Coverage fix on the automation side so this stays current going forward: kubermatic/kubelb PR follows.


Waseem826 (approved):
/approve

PR #2256, opened 2026-08-10
docs(kubelb): drop the AI multi-tenant layer, keep the agentgateway addon

ahmedwaleedmalik:
Virtual keys, budgets, spend metering and showback are not shipping in KubeLB, so the docs for them go. Only `kubelb/main` is touched; the frozen v1.2 to v1.4 pages stay as they are.

**Tutorials**

The two-page AI chapter collapses into one page for what remains: enabling the agentgateway addon, authoring your own Gateway, LLM backends, MCP and A2A. `/kubelb/main/tutorials/aigateway/` and `/kubelb/main/tutorials/ai/gateway/` both alias to it. `budgets-and-virtual-keys/` is deleted with no alias, since nothing on the remaining page answers what that URL promised.

**Elsewhere**

| Page | Change |
|---|---|
| Check catalog | drops KLB004-KLB009 and KLB019 plus the `AI` and `SpendData` capabilities; 15 checks now, IDs not renumbered |
| Insights | two sentences that listed budgets and the AI gateway as inputs |
| Air-gap install | the `images-valkey.txt` and `images-ratelimit.txt` rows, both files are gone from the release bundle |
| CE/EE matrix | the AI row is now the agentgateway addon, available in both editions |
| Dashboard | the read-only AI gateway view is no longer described as Enterprise-only, since it is gated on the addon's CRD, not on the edition |

Built locally with `hugo`; both aliases resolve to the new page.


Waseem826 (approved):
/approve

PR #2257, opened 2026-08-10
docs(kubelb): document tenant isolation in the architecture section

ahmedwaleedmalik:
The architecture page claimed that "tenants have no access to native Kubernetes resources in the management cluster" and left it there. Nothing said what enforces it, so there was no answer for the question customers actually ask, which is what a tenant credential can reach and what stops one tenant claiming another's hostname.


Waseem826 (approved):
/approve

PR #2259, opened 2026-08-12
KubeLB: document the GatewayClassNotAccepted warning event

ahmedwaleedmalik:
Documents the `GatewayClassNotAccepted` warning event the KubeLB CCM emits when a Gateway names a GatewayClass it does not serve (kubermatic/kubelb#568): when it fires, when KubeLB stays silent, and why no status conditions are written on the Gateway.


Waseem826 (approved):
/approve

PR #2264, opened 2026-08-17
Fix empty config blocks in the KDP 1.1 quickstart

mstruebing:
The nine `readfile` includes in the KDP 1.1 quickstart omit the version segment, so they resolve to `developer-platform/setup/quickstart/data/…`, which exists neither at the project root nor under `contentDir`. Hugo returns nothing, so **every config block on the page renders empty** — that is what is served today:

```
$ curl -s https://docs.kubermatic.com/developer-platform/v1.1/setup/quickstart/ | grep -c staticClients
0
```

This points the includes at the version directory that actually holds the files, restoring the ClusterIssuer, Gateway, HTTPRoute, Dex, kcp, KDP, dashboard and AI Agent values.

Verified with the CI build (`hack/ci/verify-hugo.sh`): `staticClients` and `gatewayClassName` now render on the page, where both were absent before.

`v1.0` and `v0.9.0` have the same bug (9 includes each) and are deliberately left out of this PR.

### DIFF
--- a/content/developer-platform/v1.1/setup/quickstart/_index.en.md
+++ b/content/developer-platform/v1.1/setup/quickstart/_index.en.md
@@ -59,7 +59,7 @@ This automates the process of obtaining and renewing TLS certificates from Let's
 Save the following content to a file named `cluster-issuer.yaml`, and change the value of the `email` field to your email address:
 
 ```yaml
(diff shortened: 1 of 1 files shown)


wojciech12 (approved):
/lgtm
/approve

PR #2269, opened 2026-08-18
KubeLB v1.5: measured operational caveats from release testing

ahmedwaleedmalik:
Adds the operational caveats found during v1.5 round-3 release testing to nine KubeLB pages, in both the `main` and `v1.5` trees.

These were originally #2261, closed to fold into the full v1.5 docs push. #2268 landed as a straight snapshot of `main`, so none of it made it in.

## What

| Page | Added |
|---|---|
| `tutorials/mtls-backend-transport` | The `Direct` → `MTLS` flip is fleet-wide with no per-tenant canary, costs ~15s to a few minutes of hard failure per tenant (not deterministic, L4 recovers last), and destroys client keepalive pools. `Direct` is close to hitless. `TenantState` conditions go green before traffic flows, so verify with a real request |
| `installation/upgrade` | One-time L7 disruption window, and the rollback constraints: do not downgrade the Gateway API CRDs, CCM rollback to v1.4.x needs `installGatewayAPICRDs=false`, older-chart `helm install` needs `--skip-crds` |
| `installation/management-cluster` | Envoy Gateway's stock 100 header / 60 KiB ceiling rejects with `431` at the EG hop, before KubeLB's own limits apply |
| `dashboard` | Envoy's 15s default request timeout kills watch streams on the default EG path (dashboard falls back to polling silently); auto-discovery only finds Prometheus on `:9090` |
| `security` | Dashboard images are signed from a different repository, so they need their own cosign identity |
| `tutorials/envoy-proxy/health-checks` | No fail-open (panic threshold is 0 by design), gRPC checks need HTTP/2 backends, pod-level detection needs `externalTrafficPolicy: Local`, connect-only checks never eject a backend serving errors |
| `tutorials/gatewayapi` | Cloud provider annotations must go on `spec.infrastructure.annotations`; `defaultAnnotations` on `Config` do not reach the Service Envoy Gateway provisions. Failure surfaces as `AddressNotAssigned` with the real cause only in the management cluster's Service events |
| `tutorials/loadbalancer` | `sessionAffinityConfig.clientIP.timeoutSeconds` is silently ignored, and affinity keys on the KubeLB Envoy pod IP rather than the original client |
| `cli/tunneling` | `no hostname configurable` rejection with `dns.allowExplicitHostnames: false`, and the `request-wildcard-domain` annotation to get a generated hostname |

### DIFF
--- a/content/kubelb/main/cli/tunneling/_index.en.md
+++ b/content/kubelb/main/cli/tunneling/_index.en.md
@@ -115,4 +115,8 @@ kubelb tunnel connect my-app --port 1313
 
 This will connect to the tunnel and forward traffic to the port `1313` on the local machine.
 
+### Tunnel Hostnames
+
+When the tenant has `dns.allowExplicitHostnames: false` (see [DNS](../../tutorials/security/dns/#enable-dns-automation)), a `Tunnel` with `spec.hostname` set is rejected with `no hostname configurable`. Request a generated hostname under the tenant's wildcard domain instead by annotating the `Tunnel` with `kubelb.k8c.io/request-wildcard-domain: "true"` — the same annotation used on [LoadBalancer resources](../../tutorials/loadbalancer/#load-balancer-hostname-support).
+
 For deleting, inspecting, and listing tunnels, see the [Tunnel API](../../references/api/tunnel/) documentation.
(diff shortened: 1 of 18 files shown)


Waseem826 (approved):
/approve

PR #2270, opened 2026-08-18
Drop unverified v1.5 timing and rollback claims from the KubeLB docs

ahmedwaleedmalik:
Removes the unverified measurements added in #2269, from both the `main` and `v1.5` trees. Nothing else in that PR is touched.


Waseem826 (approved):
/approve

PR #2273, opened 2026-08-19
Document addonParams in the credentials file guide

steled:
Documents the new `addonParams` key in the credentials file guide (`content/kubeone/main/guides/credentials/_index.md`), added by [kubeone#4206](https://github.com/kubermatic/kubeone/pull/4206).

Mirrors the structure/tone of the existing `customSecrets` section right above it: what the key is for, an example using the built-in `backups-restic` addon's `resticPassword`, and a note about the error raised when the same param is set both in `credentials.yaml` and in the KubeOneCluster manifest.

Not merge-blocking on kubeone#4206 being merged first, but the content assumes that PR (or an equivalent) lands.

### DIFF
--- a/content/kubeone/main/guides/credentials/_index.md
+++ b/content/kubeone/main/guides/credentials/_index.md
@@ -320,6 +320,45 @@ stringData:
   token: "{{ .CustomCredentials.MY_APP_TOKEN }}"
 ```
 
+### addonParams
+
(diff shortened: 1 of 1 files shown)


kron4eg:
/approve
/lgtm

/hold

holding it until https://github.com/kubermatic/kubeone/pull/4206 is merged.

PR #2276, opened 2026-08-20
Note fixed OSM versions in the Flatcar known issue

buraksekili:
**What this PR does / why we need it**:

Updates the Flatcar Stable 4593.2.0 known issue (kubermatic/operating-system-manager#589) in the main, v2.30 and v2.29 trees: the fix landed in OSM v1.9.1, v1.10.5 and newer (including all v1.11.x), so the "fix is in progress" note is replaced

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:


archups (approved):
/lgtm
/approve

PR #2282, opened 2026-09-01
Add Kubermatic Virtualization firewall rules how-to

mihiragrawal:
**What this PR documents**:

Adds a "Firewall Rules" how-to for Kubermatic Virtualization, covering the dashboard
workflow for firewall rules: creating, editing and deleting rules from the Firewalls
page, the per-VM firewall switch on the Networking tab, port ranges, and the relaxed
enforcement mode that keeps ICMP (ping) and DHCP working on a firewalled VM. Also
documents the `<vm>-default-deny` baseline policy and the `<vm>-allow-dns` companion
policy that accompanies a target's first outbound rule.

This also introduces a `tutorials-howtos` section for Kubermatic Virtualization,
mirroring the existing KKP layout, and files the new page under it.

**Which change does this document**:

Documents the dashboard feature added in
kubermatic/kubermatic-virtualization#439 ("Add firewall rule create, edit and per-VM posture"),

**How does this affect users**:

Firewall rules were previously visible in the dashboard but read-only. Users can now
manage them from the UI. Rules are additive and allow-only, so blocking is expressed by
turning the firewall on and then permitting what you need; and the default relaxed
enforcement mode keeps ping and DHCP alive because Kubernetes NetworkPolicy cannot
express ICMP at all.

### DIFF
--- a/content/kubermatic-virtualization/main/tutorials-howtos/_index.en.md
+++ b/content/kubermatic-virtualization/main/tutorials-howtos/_index.en.md
@@ -1,0 +1,10 @@
++++
+title = "Tutorials & How-tos"
+date = 2026-08-26T10:00:00+00:00
+weight = 6
++++
+
+Step-by-step guides for day-to-day work with Kubermatic Virtualization. Each page walks
+through a complete task from the dashboard, rather than describing a single setting in
+isolation. For the full list of settings, see the
+[Configuration Reference]({{< ref "../configuration-reference" >}}).
--- a/content/kubermatic-virtualization/main/tutorials-howtos/firewall-rules/_index.en.md
+++ b/content/kubermatic-virtualization/main/tutorials-howtos/firewall-rules/_index.en.md
@@ -1,0 +1,208 @@
++++
+title = "Firewall Rules"
+date = 2026-08-25T00:00:00+00:00
+weight = 1
++++
+
+This guide explains how firewall rules control network access to and from virtual
+machines in Kubermatic Virtualization, and how to manage them from the dashboard.
+
+## Overview
+
+A firewall rule controls which traffic is allowed to reach a virtual machine, and
+which traffic that machine is allowed to send. Each rule names:
+
+* a **direction**, inbound or outbound,
+* a **target**, the VM or VM pool the rule applies to,
+* a **source** (inbound) or **destination** (outbound),
+* the **ports and protocols** the rule permits.
+
+Rules are managed from **Firewalls** in the dashboard. Each VM also shows the rules
+that apply to it on its own **Networking** tab.
+
+Under the hood a firewall rule is a Kubernetes `NetworkPolicy` in the same namespace
+as the VM. Kubermatic Virtualization labels the policies it manages with
+`kubevirt-manager.io/managed=true`, and the Firewalls list shows only those. Write a
+policy by hand and the cluster still enforces it, but it stays out of the list until it
+carries that label. Complex policies that do carry the label are listed, and
+[Rules the form cannot express](#rules-the-form-cannot-express) covers what happens then.
+
+## How rules combine
+
+Two properties of firewall rules matter more than any individual setting:
+
+**Rules only ever allow traffic.** There is no "deny" rule. You cannot write a rule
+that blocks one particular address while leaving everything else open.
+
+**Rules are additive.** When several rules target the same VM, traffic is permitted
+if *any* of them allows it. Adding a rule can only ever open more traffic, never less.
+
+To block traffic, turn the firewall on. That closes the VM to **inbound** traffic by
+default, and you then add rules for whatever you want to permit. Nothing about the switch
+restricts what the VM sends; for that, see
+[Outbound rules and DNS](#outbound-rules-and-dns).
+
+## Turning a VM's firewall on and off
+
+Open the VM, go to the **Networking** tab, and use the **Firewall** switch.
+
+* **Off**: the VM accepts all inbound traffic. This is the default for a new VM.
+* **On**: inbound traffic is blocked unless one of the VM's rules allows it.
+* **Unavailable**: the switch is greyed out. The VM was not created through this
+  dashboard, so it does not carry the label firewall rules select on, and no rule can be
+  enforced against it. See [Which VMs can be targeted](#which-vms-can-be-targeted).
+
+Turning the firewall on creates a policy named `<vm-name>-vm-default-deny`. The `vm` part
+is in the name because a VM and a VM pool can share a name in one namespace, and each type
+needs its own system policy. For a pool the name is `<pool-name>-pool-default-deny`.
+
+The policy appears in the Firewalls list marked **Default deny**, and it is managed for
+you, so the actions menu offers **View YAML** but not **Edit** or **Edit YAML**. Turning
+the switch back off removes it. So does deleting it from the Firewalls list, and the
+delete dialog says which outcome to expect. Either the VM accepts all inbound traffic
+again, or it stays closed because other inbound rules still apply.
+
+The switch affects **inbound** traffic only. The VM's own outbound traffic stays
+unrestricted until you add an outbound rule.
+
+{{% notice note %}}
+Adding an inbound rule also closes the VM, even with the switch off. That is how
+Kubernetes network policies work. The switch is there to make the VM's posture explicit
+and visible, instead of leaving it to depend on whether some rule happens to exist. When
+a VM is closed only by its rules, the Networking tab says so, and deleting the last rule
+reopens it.
+{{% /notice %}}
+
+## Creating a rule
+
+Go to **Firewalls** and choose **New Firewall Rule**.
+
+| Field | What it does |
+| --- | --- |
+| **Name** | Identifies the rule. Lowercase letters, numbers and hyphens. Cannot be changed later. |
+| **Direction** | **Inbound** controls traffic arriving at the target. **Outbound** controls traffic the target sends. |
+| **Target Type** / **Target** | The virtual machine or VM pool the rule applies to: the receiver for an inbound rule, the sender for an outbound one. |
+| **Source** / **Destination** | Where the traffic comes from or goes to. See below. |
+| **All ports** | Allow every port for the chosen peer, instead of listing them. |
+| **From port** / **To port** / **Protocol** | The ports to permit. Leave **To port** empty for a single port, or set it to allow a range. |
+| **Strict enforcement** | See [Ping and DHCP](#ping-and-dhcp). Leave this off unless you need full isolation. |
+| **Labels** | Optional labels for your own grouping. |
+
+The source or destination can be:
+
+* **IP range**: a CIDR such as `10.0.0.0/8` or `2001:db8::/32`.
+* **Virtual Machine** or **VM Pool**: another VM or pool in the same namespace.
+* **Namespace**: every workload in the named namespace.
+* **Anywhere**: any address. Use this to open a port to the world.
+
+The first rule of a given direction for a target gets a warning. The form says the
+target *"allows all inbound traffic today"* (or outbound), and that after saving,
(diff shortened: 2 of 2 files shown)


content/kubermatic-virtualization/main/tutorials-howtos/firewall-rules/_index.en.md
```diff
@@ -0,0 +1,208 @@
...
+{{% notice warning %}}
+Editing a rule replaces the whole policy. Changes made outside the dashboard to the same
+rule, such as extra ports or additional sources, are lost when you save the form. If a rule is
+too complex for the form to represent, the dashboard says so and opens the YAML editor
+instead of showing you a simplified version it would then destroy.
+
+The YAML editor behaves differently. It submits the exact version you opened, so if the
+rule changed underneath you the save is refused with *"This rule changed since you opened
+it — reload and reapply your edit"* instead of overwriting. The form carries no such
+check, which is why concurrent edits are lost there but rejected here.
```
Waseem826, line 169: This is the other way round. The form PUTs the object it fetched when Edit was clicked, `resourceVersion` included (`routes/firewalls.tsx:176-195`), and the update mutation maps a 409 to exactly the message quoted here (`hooks/mutations/use-service-mutations.ts:366-368`). The YAML dialog has its own conflict text with a "Load latest version" button (`components/yaml/resource-yaml-dialog.tsx:81-84`). Suggest rewriting so both editors are described as rejecting a stale save, with the quoted string attributed to the form.

content/kubermatic-virtualization/main/tutorials-howtos/firewall-rules/_index.en.md
```diff
@@ -0,0 +1,208 @@
...
+
+Deleting tells you what the change will do before you confirm. For example, that a VM
+will accept all inbound traffic again because you are removing its last inbound rule, or
+that a system DNS allowance is being removed alongside the outbound rule that needed it.
+
+{{% notice warning %}}
+Editing a rule replaces the whole policy. Changes made outside the dashboard to the same
+rule, such as extra ports or additional sources, are lost when you save the form. If a rule is
+too complex for the form to represent, the dashboard says so and opens the YAML editor
+instead of showing you a simplified version it would then destroy.
```
Waseem826, line 164: The form does not silently drop anything. `parse-network-policy.ts` loads every port row (the 8-row cap only hides the Add button) and refuses a rule with more than one peer, which is what sends it to the YAML editor; labels and annotations are merged from the live object in `handleEdit`. Suggest dropping the "extra ports or additional sources ... are lost" sentence and keeping the rest of the paragraph, which is accurate.

content/kubermatic-virtualization/main/tutorials-howtos/firewall-rules/_index.en.md
```diff
@@ -0,0 +1,208 @@
...
+```bash
+kubectl -n <namespace> get networkpolicy <rule-name> -o yaml
+```
+
+Nothing rejects a policy that matches no VM. It is created cleanly and then does nothing,
+so the form tells you while you are still filling it in, underneath the **Target** field:
+
+* *"Applies to 1 virtual machine."* means the rule will be enforced. The number is a real
+  count, so a pool rule reports how many machines it selects, for example *"Applies to 4
+  virtual machines."*
```
Waseem826, line 197: The match line only renders when the target type is Virtual Machine, and the count is `targetableVms.includes(target) ? 1 : 0` (`components/forms/firewall-form.tsx:210-215`). Pool targets get no match feedback at all, so "Applies to 4 virtual machines" cannot appear. Suggest removing the pool example and saying the line appears for VM targets only.

content/kubermatic-virtualization/main/tutorials-howtos/firewall-rules/_index.en.md
```diff
@@ -0,0 +1,208 @@
...
+
+* **Off**: the VM accepts all inbound traffic. This is the default for a new VM.
+* **On**: inbound traffic is blocked unless one of the VM's rules allows it.
+* **Unavailable**: the switch is greyed out. The VM was not created through this
+  dashboard, so it does not carry the label firewall rules select on, and no rule can be
+  enforced against it. See [Which VMs can be targeted](#which-vms-can-be-targeted).
+
+Turning the firewall on creates a policy named `<vm-name>-vm-default-deny`. The `vm` part
+is in the name because a VM and a VM pool can share a name in one namespace, and each type
+needs its own system policy. For a pool the name is `<pool-name>-pool-default-deny`.
```
Waseem826, line 57: `buildBaselinePolicy` has a single caller, hard-coded to `'VirtualMachine'` (`components/vm-detail/vm-firewall-section.tsx:54`), and there is no firewall section under pool-detail. The `-pool-default-deny` name exists in the naming scheme but nothing in the dashboard creates it. Suggest saying the switch is VM-only today and pools are closed only by their inbound rules.

content/kubermatic-virtualization/main/tutorials-howtos/firewall-rules/_index.en.md
```diff
@@ -0,0 +1,208 @@
...
+* *"Matches no virtual machine — this rule will have no effect until one exists."* means no VM of
+  that name exists in the namespace yet. Check the name for a typo.
+* *"This virtual machine cannot be targeted by a firewall rule: it was not created here, so it
+  does not carry the label rules select on."* See below.
+
+### Which VMs can be targeted
+
+A firewall rule selects VMs by a label that this dashboard applies when it creates a VM.
+Nothing back-fills that label, so a VM created another way, whether by `kubectl`, a YAML
+manifest, or an import, cannot be targeted by a rule. The API server accepts such a rule and it then
```
Waseem826, line 207: Naming the label would give the reader a remedy. The selector is `fw.kubevirt-manager.io/vm-name=<vm-name>` on `spec.template.metadata.labels` (`lib/constants.ts:20`, `hooks/use-vm-names.ts:41-42`), so an imported VM can be made targetable by adding it. Suggest naming the label and where it goes.

content/kubermatic-virtualization/main/tutorials-howtos/firewall-rules/_index.en.md
```diff
@@ -0,0 +1,208 @@
...
+
+For ports that are not contiguous, or that use a different protocol, add more rows. Eight
+is the limit. Once you reach it the button to add another disappears, so put any further
+ports in a second rule against the same target, which works because rules are additive.
+
+## Ping and DHCP
+
+By default, rules are created in a relaxed enforcement mode that keeps **ICMP (ping)
+and DHCP working** on a firewalled VM, no matter what the rules say. This is almost
+always what you want:
```
Waseem826, line 117: The relaxed mode is the Kube-OVN annotation `ovn.kubernetes.io/network_policy_enforcement: lax` on the policy, and strict is simply its absence (`components/forms/build-network-policy.ts:17-18,115`). Since the troubleshooting section sends people to the policy YAML, they have no way to recognise the mode without this. Suggest naming the annotation once and noting the ping/DHCP allowance is a Kube-OVN behaviour.

content/kubermatic-virtualization/main/tutorials-howtos/firewall-rules/_index.en.md
```diff
@@ -0,0 +1,208 @@
...
+reopens it.
+{{% /notice %}}
+
+## Creating a rule
+
+Go to **Firewalls** and choose **New Firewall Rule**.
+
+| Field | What it does |
+| --- | --- |
+| **Name** | Identifies the rule. Lowercase letters, numbers and hyphens. Cannot be changed later. |
```
Waseem826, line 82: Names ending in `-default-deny` or `-allow-dns` are rejected on create (`components/forms/firewall-form-schema.ts:109-117`). Since this page introduces both suffixes it is worth saying they are reserved.

content/kubermatic-virtualization/main/tutorials-howtos/_index.en.md
```diff
@@ -0,0 +1,10 @@
++++
+title = "Tutorials & How-tos"
+date = 2026-08-26T10:00:00+00:00
+weight = 6
```
Waseem826, line 4: Three sibling sections already use `weight = 6` (`access-control-guide`, `configuration-guide`, `references`), so where this lands in the sidebar is tie-broken rather than chosen. Suggest picking a free weight.

Waseem826 (commented):
Some feedback, PTAL

content/kubermatic-virtualization/main/tutorials-howtos/_index.en.md
```diff
@@ -0,0 +1,10 @@
...
++++
+
+Step-by-step guides for day-to-day work with Kubermatic Virtualization. Each page walks
+through a complete task from the dashboard, rather than describing a single setting in
+isolation. For the full list of settings, see the
+[Configuration Reference]({{< ref "../configuration-reference" >}}).
```
Waseem826, line 10: The KKP `tutorials-howtos/_index.en.md` this mirrors ends with a `## Pages` heading and `{{% children depth=5 %}}`, so the section landing page lists its how-tos. Without it this page shows only the intro prose. Suggest adding the same listing.

content/kubermatic-virtualization/main/tutorials-howtos/firewall-rules/_index.en.md
mihiragrawal: You're right, I had it backwards. Rewritten: both editors are now described as working from the version they fetched when the rule was opened, so a rule changed in the meantime is never silently overwritten. The quoted string is attributed to the form, and the YAML dialog's **Load latest version** button is described separately.

content/kubermatic-virtualization/main/tutorials-howtos/firewall-rules/_index.en.md
mihiragrawal: Right — dropped that sentence.

content/kubermatic-virtualization/main/tutorials-howtos/firewall-rules/_index.en.md
mihiragrawal: Removed the pool example; the line is described as VM-only now.

content/kubermatic-virtualization/main/tutorials-howtos/firewall-rules/_index.en.md
mihiragrawal: Took the pool naming out. Happy to add an explicit line that a pool has no switch if you'd prefer it spelled out.

content/kubermatic-virtualization/main/tutorials-howtos/firewall-rules/_index.en.md
mihiragrawal: Named it, with a kubectl patch for VMs that came from elsewhere.

content/kubermatic-virtualization/main/tutorials-howtos/firewall-rules/_index.en.md
mihiragrawal: Added the annotation, plus a note that it's Kube-OVN behaviour and that --np-enforcement sets the default.

content/kubermatic-virtualization/main/tutorials-howtos/firewall-rules/_index.en.md
mihiragrawal: Added to the Name row.

content/kubermatic-virtualization/main/tutorials-howtos/_index.en.md
mihiragrawal: Moved it to weight = 4.

content/kubermatic-virtualization/main/tutorials-howtos/_index.en.md
mihiragrawal: Added, same as the KKP page.

PR #2284, opened 2026-09-08
Fix component versions for Kubermatic Virtualization

mihiragrawal:
Fix component versions for Kubermatic Virtualization

The v1.2.0 component table was a stale copy of an earlier release; the v1.1.0
and main tables were only partially updated. Correct all three against the
versions pinned in pkg/kubeone/types.go at the matching release tag, verified
against a running cluster where the component is actually deployed.


mohamed-rafraf (approved):
/approve 
/lgtm

PR #2285, opened 2026-09-08
Consolidate KubeV configuration reference and fix stale documentation

mihiragrawal:
configuration-reference.md was published without Hugo frontmatter, was not linked from any page, and duplicated roughly 96% of configuration-guide. Delete it across v1.1.0, v1.2.0 and main, and carry over the content that only existed there.

The Kube-OVN architecture section is migrated with a correction: it described inter-node traffic as travelling through encrypted Geneve tunnels. Kube-OVN's IPsec support is opt-in and KubeV deploys with --enable-ovn-ipsec=false, so a
note now states that tunnel traffic is not encrypted.

The air-gapped preparation steps gain a Helm chart mirroring step. The deleted page referenced a "kubev mirror-charts" command that does not exist; the new step documents helm pull/push against the chart registry instead. Offline installs need the charts mirrored and the guide never said so. The note claiming platform images need separate registry  credentials is removed: offline mode configures containerd with a wildcard mirror, so the single containerRegistry covers every image pull.

Document copying the installer binary into PATH as kubev, since the documentation invokes the CLI by that name while the downloader installs it as kubermatic-virtualization.

Fix stale values that were copied forward into the v1.2.0 directory:

  - the install page advertised v1.1.0 as the latest release, passed $USERNAME/$PASSWORD to the downloader instead of the KUBEV_ prefixed variables set earlier on the same page, and dropped the ./ prefix from the downloader invocations
  - the declarative installation sample showed nodes at Kubernetes v1.33.0 where all three releases ship v1.34.7
  - the CRD reference linked to the Kubernetes v1.25 API documentation


content/kubermatic-virtualization/main/getting-kubermatic-virtualization/_index.en.md
```diff
@@ -53,18 +53,34 @@ Once your credentials are set, run the downloader to retrieve and install the la
...
 INFO[2026-04-23T14:24:06+01:00] Downloading binary layer                      media_type=application/octet-stream size="79.15 MB"
 INFO[2026-04-23T14:24:06+01:00] Download complete                             path=./kubermatic-virtualization
 ```
 
 Once complete, the `kubermatic-virtualization` binary is ready to use in your current directory.
 
+### Making the CLI Available as `kubev`
+
+The downloader installs the binary under its full name, `kubermatic-virtualization`. Throughout this
+documentation the CLI is invoked as `kubev`, which is also the name the binary reports in its own
```
toschneck: I don't think we should do this, please update the documentation cli with the real name

toschneck (changes requested):
@mihiragrawal please don't use something like this, as you never know where a user start to read the docs

content/kubermatic-virtualization/main/getting-kubermatic-virtualization/_index.en.md
mihiragrawal: Fixed in e0a5453

